PR #1, opened 2016-06-21
Modularize

callahad:
The end goal is to split the library up into smaller modules with consistent interfaces, so we can isolate and better reason about interactions. This PR just moves lines of code into different files, without changing any logic, but it starts us down that path.

Maybe something like this?

```
├── main
├── lib
├── handlers
│   ├── openid_provider
│   └── website
└── providers
    ├── gmail
    └── smtp
```

I'd like to design the providers such that you can interrogate them, and they can respond with whether or not they can handle a given email address. E.g., `gmail::can_auth("example@gmail.com")` would be `true`. Then we just just put them in priority order and use the first to say "yes."


djc:
Hmm, not sure this is the right way to split up the code. I think it would be better to start putting, say, the storage code in a separate module, so that most of the code doesn't have to know about Redis. Similarly, there could be a token module that can take care of the JWS creation/validation, which can also abstract away the crypto library/algorithms used.

Also, the code that's there is not just gmail; it supports hotmail.com just fine provided you provide configuration for a live.com-authorized application.

Not sure yet if the direction you propose for `can_auth()` makes sense to me; I'm wondering how it can intersect with the way at least the famous IdPs get configured. It probably still does make sense to extract the two different authentication flows into separate modules.

Another more useful abstraction would probably be to have a kind of URL builder utility method or class.

callahad:
> putting, say, the storage code in a separate module, so that most of the code doesn't have to know about Redis.

Yes, I want that too :)

> there could be a token module that can take care of the JWS creation/validation, which can also abstract away the crypto library/algorithms used.

And that!

> the code that's there is not just gmail; it supports hotmail.com just fine

We may be able to extend a generic OAuth2 client to support Gmail, but we will need Gmail-specific logic on day one to handle Gmail's `+`-addressing and `.`-insensitivity, since Gmail will only return the canonical address from its federated endpoints, but the user may have entered a custom `.` pattern or subaddress on a specific RP. Also, it has to handle the interchangeability of `gmail.com` and `googlemail.com`.

...so we need some kind of isolated place to put that kind of stuff.

Lastly, we may yet determine a reliable way to detect Google Apps for your Domain accounts, which may require DNS probes, etc. that would be completely inapplicable to other providers.

> Not sure yet if the direction you propose for can_auth() makes sense to me; I'm wondering how it can intersect with the way at least the famous IdPs get configured.

I believe we should eventually alter the configuration to work per-provider/protocol, not per domain, since a single provider may be able to handle many domains, some of which may only be discoverable at runtime. I'm imagining the configuration calling some kind of constructor that then returns a struct with `can_auth()`, `do_auth()`, etc. on it. So you might have something like `let providers = vec![gmail::new(API_KEY, API_SECRET), mslive::new(...), mailinator::new(), smtp::new(...)]`, and could re-order, omit, or swap out providers as desired.

djc:
So do you want to take a whack at trying things like extracting the store and/or token code, or shall I see what happens? Your current attempt at modularizing seems a bit random to me. I'd prefer to see how it plays out if the main flow is more concise because helpers are factored out.

In general, it seems like I'm less a fan of spreading code out over multiple files than you. In tightly coupled code, it seems to me that having to switch between files is just harder than keeping the code together; and this is, as I experience it, a very large file.

callahad:
It's a bit random, yes. Trying to just isolate things along function lines first, and then pull out commonalities. We're on different timezones again, so we could alternate on pushing to a branch?

callahad:
As to the one-big-file notion, I'm a huge fan of deliberate separation of concerns. We could merge everything back into a single big file by using inline `mod foo { ... }` statements, but I find separate files easier for focusing attention and preserving encapsulation.

djc:
I've pushed a bunch of stuff that should make things more modular (cribbing your nice `router!()` macro use and fixing clippy lints in the process, with your name on it). Let me  know what you think. I'm happy to rebase your further changes if that makes sense.

callahad:
Closing this; no longer useful.

PR #2, opened 2016-06-24
Use docopt for CLI arg parsing

callahad:
I also trimmed down the comments in main.rs, since things seem sufficiently declarative now that we've got the `router!` macro in place.


callahad:
@djc how's this?

djc:
Would you mind splitting in three commits?

1. Cleaning up commits
2. Implementing docopt
3. Configurability of address and port

Also can you explain this version contraption?

```
.and_then(|d| d.version(Some(VERSION.to_string())).decode())
```

PR #13, opened 2016-08-28
add RP nonce

onli:
See https://github.com/onli/sinatra-browserid/commit/1dc22b29c15304db34b041ab0be01eeaa2ae322d for example usage

### DIFF
--- a/src/email.rs
+++ b/src/email.rs
@@ -44,11 +44,13 @@ pub fn request(app: &AppConfig, params: &QueryMap) -> Value {
     // Store data for this request in Redis, to reference when user uses
     // the generated link.
     let client_id = &params.get("client_id").unwrap()[0];
+    let nonce = &params.get("nonce").unwrap()[0];
     let session = session_id(&email_addr, client_id);
     let key = format!("session:{}", session);
     let res = app.store.store_session(&key, &[
         ("email", &email_addr.to_string()),
         ("client_id", client_id),
+        ("nonce", nonce),
         ("code", &chars),
         ("redirect", &params.get("redirect_uri").unwrap()[0]),
     ]);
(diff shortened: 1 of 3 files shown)


stephank:
The example creates the nonce at form render, and I suppose its lifespan is tied to the session? I'm wondering if we should account for long timespans between form render and form submit, or if that's something the RP should just handle by giving the user a generic error and letting him/her try again.

stephank:
I'm thinking, our `/auth` endpoint currently does `POST` only, whereas the OAuth2 spec says `POST` is optional but `GET` is required.

If we add support for `GET /auth`, with the parameters in the query string, that means we can change to RP form to submit to the RP itself, which then generates to nonce and does a 303 redirect to `/auth` with the parameters.

It's a tradeoff: shorter nonce lifespan but an extra roundtrip. I'm not sure how beneficial that'd be, actually. The little extra bonus is that we're more compliant with the OAuth2 spec if we support `GET`.

stephank:
Regardless of whether `GET` and #16 is good or bad, this PR works fine.

The only comment I have is that `nonce` is now required. If the intention was for it to be optional, it'd need a small adjustment.

onli:
Thanks for looking at that, @stephank !

> The example creates the nonce at form render, and I suppose its lifespan is tied to the session?

Yes, well, it depends on how the RP handles it. I don't think that is something for the broker to decide. In the sinatra module I'm saving it in the session and it thus can live very long.

> The only comment I have is that nonce is now required. If the intention was for it to be optional, it'd need a small adjustment.

That was the intention, however not something I thought very much about. Is there a big drawback making the nonce required?

djc:
I think we should require the nonce, per the OIDC spec on Implicit Flow.

It would be better, IMO, if RP forms should submit to the RP server, which should then create the nonce only for submitted forms and redirect the user agent to the broker for the auth request.

onli:
What's the difference? Nonces are tied to sessions anyway.

But the RP can handle it as he wants. He can opt to generate the nonce on submit, or on form creation. It does not change the broker side (or this PR).

stephank:
The small benefit of putting the RP in between and generating the nonce on submit is that the RP can delay session creation. I was specifically thinking sites that have the login form permanently on every page, such as a JS widget. In that case, you can still cache the entire page.

Regardless, the PR works great! 👍

onli:
Okay, that was enough confirmation to me. No need stopping discussion ;) But I felt it's time to merge this.

Thanks @stephank !

PR #14, opened 2016-09-26
Configurable SMTP host.

stephank:
This is useful when testing locally with [MailHog], for example.

 [MailHog]: https://github.com/mailhog/MailHog


djc:
Good stuff, thank you!

Issue #11, opened 2016-08-19
Check handling of GMail address variants


djc:
With dots and pluses. Ideally, it should not canonicalize?

callahad:
For gmail, we should normalize. Here's what we did in Persona: https://github.com/mozilla/persona-gmail-bridge/blob/master/lib/email.js

There are some gross things in there that we could have avoided with more mature libraries for parsing email addresses, but it worked well enough.

callahad:
(Meaning, we should canonicalize in that manner to see if there's a match between user input and federated return value. if so, we should then return to the RP an identical string to the user's input)

djc:
Right. This means changing the code that compares the JWT subject and the auth request email address to use a comparison function that canonicalizes email addresses.

skorokithakis:
This seems simple enough that I can give it a go, I'll try it tonight and see how far I get.

djc:
@skorokithakis awesome! Let me know if you run into trouble, I'm happy to assist.

skorokithakis:
@djc Will do, thanks!

PR #18, opened 2016-10-05
canonicalize gmail address (#11)

onli:
This follows pretty closely https://github.com/mozilla/persona-gmail-bridge/blob/master/lib/email.js, with some adjustments around the borrow checker which I don't master yet.

I tested the code in question, but not the whole flow (missing a good test environment).

### DIFF
--- a/src/oidc.rs
+++ b/src/oidc.rs
@@ -73,6 +73,34 @@ pub fn request(app: &AppConfig, params: &QueryMap) -> Url {
 
 }
 
+pub fn canonicalized(email: &str) -> String {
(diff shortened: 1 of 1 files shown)


djc:
Rebased your change as 11ea3499092b7d2d6985ee53819354ab78aabe50, simplified in e9ca7535564d0fd275ffc419c1044b40cd69ffec and added tests in d0aa7c90d9f45ebb332a0bcfd8a4f8a2ed2a149a.

onli:
Thanks!

PR #19, opened 2016-10-06
Add a macro to simplify route handler functions.

stephank:
This is an attempt at fixing the routing related TODO in `main.rs`. Not sure if this is something we want, just throwing it out there. Low prio, in any case.

(Probably easier to review with `?w=1` / `-w`)


callahad:
Compiles and LGTM. Merging. Thanks :)

PR #20, opened 2016-10-07
Add back `email::request` call.

stephank:
Made a boo boo. 🙈


djc:
Haha, thanks for fixing it.

PR #22, opened 2016-10-12
Better isolate serde bits

callahad:
Mainly just to better isolate the bits that serde generates and more closely match serde's docs for supporting the stable compiler.

See the commit message on 52c16365028189f3301c1f614f8e3b4f67b2fab5 for more details.


callahad:
With @stephank's 👍, I'm merging :)

djc:
The first commit here doesn't actually build, which I think isn't great to have land.

callahad:
That was intentional -- if the rename and modifications were in the same commit, the cumulative diff was more difficult to meaningfully review.

I'm 100% on board with never merging into master unless things build, of course 😉

djc:
It could still have had the minimal other stuff to make it compile. Having commits that don't build in the history is a PITA when you later want to find a regression (e.g. by bisecting).

djc:
(Anyway, see my post to the mailing list about some further thoughts.)

callahad:
Link to mailing list thread for reference: https://groups.google.com/forum/#!msg/portier/CDlUhUdQykE/_E6AqmYREAAJ

PR #26, opened 2016-10-13
Run rustfmt with default settings. What looks bad?

callahad:
Some of the formatting choices aren't great for us, but this is our starting point.

...now to figure out which ones we care about, and how to configure rustfmt to work better for us.

**NOTE:** This is  a work-in-progress discussion; not necessarily planning to merge this branch directly!

### DIFF
--- a/src/store_cache.rs
+++ b/src/store_cache.rs
@@ -1,10 +1,7 @@
 use std::cmp::max;
 use std::io::Read;
 use super::hyper::client::Client as HttpClient;
-use super::hyper::header::{
-    CacheControl as HyCacheControl,
-    CacheDirective as HyCacheDirective
-};
+use super::hyper::header::{CacheControl as HyCacheControl, CacheDirective as HyCacheDirective};
 use super::redis::Commands;
 use super::error::{BrokerError, BrokerResult};
 use super::store::Store;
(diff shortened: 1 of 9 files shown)


callahad:
Odd. Seems like rustfmt takes a few invocations before it settles?

callahad:
Weird! It'll work just find for a few invocations, and then it'll reformat something.

If I run it enough times, it'll format it back:

screenshot from 2016-10-12 21-19-17


So.... seems like `cargo fmt` isn't deterministic, yet. Upstream bug, here I come.

callahad:
For instance, consider these three consecutive runs:

e9970cbb37d7aafc64801ff887bb0203ee505fda indents some comments
e3d709505d9fd5017a601de53074870fc1537d80 de-indents the same comments
345fc2825f1bc22ca29f1cbe4f77fc7da89be238 re-indents the same comments

WTF?

src/store_cache.rs
```diff
@@ -73,19 +70,21 @@ impl StoreCache {
 
             Ok(data)
 
-        }, |data| {
-            Ok(data)
-        })
+        },
+                           |data| Ok(data))
```
stephank: This one is especially weird. Looks like it tried to align with the `(` from `map_or_else`. Though `map_or_else` felt weird to use anyway, so I wouldn't mind if this was rewritten.

callahad:
Closing this for now - we'll revisit rustfmt post 1.0 :)

PR #27, opened 2016-10-13
Add a script to build Docker images.

stephank:
Using this script, I've pushed a first experimental image to [portier/broker](https://hub.docker.com/r/portier/broker/). The image is ~3MB compressed, and ~7MB extracted on disk.

I've tested that the oidc flow is able to communicatie over HTTPS with Google.

If we merge this, I can for now configure a private Jenkins instance to poll `master` every 15 minutes and build images. (The current image was also produced on this Jenkins instance.)

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+# This script builds the portier/broker Docker image. The image contains a
+# static release build of portier-broker, and a bundle of root certificates
+# extracted from Debian.
+set -xe
+cd "$(dirname "${0}")/"
+
+SOURCE_DIR="${PWD}"
+TARGET_DIR="target/x86_64-unknown-linux-musl/release"
+DOCKER_SOCK="/var/run/docker.sock"
+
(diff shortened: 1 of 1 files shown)


djc:
I'm not sure I'd be a good reviewer for this. @callahad, do you have experience with this stuff?

buro9:
This looks fine.

The only thing missing that I'd be tempted to add is a `USER nobody` line to the output Dockerfile if we don't actually need to run portier broker as a superuser within the container. Do we need to?

djc:
So far I've always run it behind a proxying web server. I'm not sure how easy that is to set up with Docker stuff?

stephank:
If your proxy runs on the host or elsewhere, you expose the port on the host. If your proxy is a container, you link the two containers, and Docker provides a resolvable hostname you can configure in your proxy.

I've rebased added the `nobody` line. :)

stephank:
Thanks! I've setup Jenkins to track `master` now. :)

Issue #5, opened 2016-06-25
Measure (and increase) test coverage


djc:

callahad:
We want to make sure we don't panic on invalid input data, a la #28

callahad:
Potentially helpful blog post on generating code coverage reports for Rust: http://sunjay.ca/2016/07/25/rust-code-coverage

callahad:
Also this: https://users.rust-lang.org/t/tutorial-how-to-collect-test-coverages-for-rust-project/650/33

PR #29, opened 2016-10-14
Prevent panic on invalid email address. Fixes #28

### LINKED ISSUE #28: Panic on invalid email address
If a request contains a `login_hint` without an `@`, the `emailaddress` crate causes a panic.

Upstream merged a pull request that fixed it, but hasn't made a release yet: https://github.com/pwoolcoc/emailaddress-rs/pull/5

I've requested the author make a release with this fix. We should upgrade when that happens.

callahad:
We should change this dep to 0.3.1 once that's released.
See: https://crates.io/crates/emailaddress

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ serde_codegen = "0.8.0"
 
 [dependencies]
 docopt = "0.6.81"
-emailaddress = "0.3.0"
+emailaddress = "0.3.1"
 env_logger = "0.3.4"
 hyper = "0.9.10"
 iron = "0.4.0"


src/lib.rs
```diff
@@ -297,3 +297,17 @@ broker_handler!(CallbackHandler, |app, req| {
         try!(oidc::verify(app, session, code))
     )
 });
+
+
+#[cfg(test)]
+mod tests {
+    // TODO: Refactor so these tests fit somewhere more natural...
```
stephank: This is more of a test of `emailaddress` itself, so I'm not sure it belongs here at all.

Tests here could call the handler functions and check the result, but that also sounds more like integration testing than unit testing. Stuff like `crypto.rs` would be good unit test candidates, I think.

callahad: Maybe the resolution here is to leave this open and unmerged until I / we can write some tests covering this sort of thing?

stephank:
`emailaddress` 0.3.1 was just published! :D

callahad:
Changed the commit to just bump the emailaddress version, added a note to test for regressions in #5

Issue #31, opened 2016-10-17
Display form allowing for manual input of code when using email-based authentication


callahad:
Clicking the link works fine, but the Broker needs to display an interstitial page that:

1. Tells the user to check their email
2. Allows them to manually type in the code

stephank:
This one was fixed in #40.

PR #40, opened 2016-10-17
Response templates and RP redirecting

### LINKED ISSUE #17: Use templates for the email message.
From a TODO in the code: https://github.com/portier/portier-broker/blob/5631a7ed81bff54776669dafd7df3c08a78519ea/src/email.rs#L63

Compile-time or run-time?

Run-time of course means more files (right now just the binary and config), but also more configurability without a rebuild. For run-time, I'm leaning towards [liquid](https://github.com/cobalt-org/liquid-rust).

But compile-time may be good enough for us, considering the primary goal is a central hosted service.

stephank:
Fixes #17 and #31. Changes:

 - Added a form to the email flow, so the one-time pad from the mail can also be entered in an input. (As an alternative to clicking the link.)
 - Created templates for the email itself, in both plain text and HTML.
 - Added a template for errors, when we cannot redirect back to the RP.
 - Implemented reporting errors to the RP through redirects, following the OAuth2 spec.

Even though both @djc and I felt strongly keeping templates simple search/replace operations, throughout implementing this I felt that actually be more complicated than just integrating a template engine. Reason is that we need escaping, and probably also partials, sections and assets down the road, if we're going to build a more full-fledged UI.

I picked Mustache for now, because Liquid doesn't support partials, and Mustache is probably the better known alternative. The Mustache implementation for Rust is also reasonably mature, as used by nickel. The way things are setup, I believe we can easily swap it out for something else if needed.

Regarding error redirects, I followed RFC 6749 4.1.2.1., but I have a suspicion we may have to follow the response mode? Right now, it does a 303 redirect, which results in a GET request with query params to the RP verify endpoint, but our response mode is always `form_post`.


stephank:
Rebased, but now again GitHub displays commits out of order. :/

callahad:
How exactly are you rebasing? Weird that it looks correct in the Git CLI, but mis-ordered on GitHub :-/

callahad:
This looks good to me, but I'd like to defer to @djc for a formal r+.

stephank:
Nothing special, just did a normal `git rebase origin/master`.

But GitHub has had this for years: https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/

stephank:
I've added commits to POST errors instead, as per https://github.com/portier/demo-rp/issues/7.

djc:
(Sent an email to GitHub support about the stupid commit order.)

src/config.rs
```diff
@@ -65,10 +67,53 @@ impl Deserialize for store::Store {
...
+    fn default_forward() -> Template {
+        Template(mustache::compile_path("tmpl/forward.mustache").unwrap())
+    }
+}
+
+
 impl AppConfig {
-    pub fn from_json_file(file_name: &str) -> Result<AppConfig, Error> {
+    /// Read a JSON configuration file.
+    pub fn from_json_file(file_name: &str) -> Result<AppConfig, ConfigError> {
```
djc: This belongs in a different commit, right?

src/config.rs
```diff
@@ -65,10 +67,53 @@ impl Deserialize for store::Store {
...
+            builder = builder.insert_str(key, value);
+        }
+        let mut out: Vec<u8> = Vec::new();
+        self.0.render_data(&mut out, &builder.build());
+        String::from_utf8(out).unwrap()
+    }
+}
+
+
+
```
djc: Nit: one too many empty newlines.

src/config.rs
```diff
@@ -65,10 +67,53 @@ impl Deserialize for store::Store {
...
+impl Templates {
+    fn default_forward() -> Template {
+        Template(mustache::compile_path("tmpl/forward.mustache").unwrap())
```
djc: It makes more sense for me to inline this into the `default_templates()` method, below.

src/serde_types.in.rs
```diff
@@ -45,4 +59,7 @@ pub struct AppConfig {
     pub token_validity: usize,
     /// Mapping of Domain -> OIDC Provider
     pub providers: HashMap<String, Provider>,
+    /// Template files
```
djc: I don't quite get what you're going for with this part. It makes sense to me to have a `Templates` struct, and I guess it also makes sense to have an instance of that struct in the `AppConfig`. However, does that necessarily mean we have to use `serde` for all this stuff? `impl Deserialize for Template` should just be a function `compile_template()` that's called once for each template in `default_templates()`.

build-docker.sh
```diff
@@ -20,12 +20,13 @@ container="$(
 )"
 trap "docker rm -f ${container}" exit
 
-docker cp "${TARGET_DIR}/portier-broker" "${container}:/"
+docker cp "${TARGET_DIR}/portier-broker" "${container}:/tmp/portier-broker"
```
djc: This looks like it should be folded into e34b1f99750b29f4ced4e85910e2edfdc9a2251c, otherwise that commit has an invalid docker file creation script, right?

src/config.rs
```diff
@@ -102,6 +102,14 @@ impl Templates {
         Template(mustache::compile_path("tmpl/confirm_email.mustache").unwrap())
     }
 
+    fn default_email_html() -> Template {
```
djc: Could you split this commit in two, please, first templatizing the text email and then adding the (also templated) HTML email variant separately?

src/lib.rs
```diff
@@ -212,11 +219,14 @@ broker_handler!(AuthHandler, |app, req| {
...
     let client_id = try_get_param!(params, "client_id");
-    let nonce = try_get_param!(params, "nonce");
     let redirect_uri = try!(
         Url::parse(try_get_param!(params, "redirect_uri"))
             .map_err(|_| BrokerError::Input("redirect_uri is not a valid URL".to_string()))
     );
+    // Per the OAuth2 spec, we may redirect to the RP once we have validated client_id and
+    // redirect_uri. In our case, this means we make redirect_uri available to error handling.
+    req.extensions.insert::<RedirectUri>(redirect_uri.clone());
+    let nonce = try_get_param!(params, "nonce");
```
djc: Nit: can this last line not move in this commit, please? And if it moves, it should probably go after the `let email_addr`, as well? Then, maybe add a blank line before it.

src/lib.rs
```diff
@@ -93,25 +93,64 @@ macro_rules! try_get_param {
...
+                BrokerError::Input(_) | BrokerError::Provider(_) => {
+                    pairs.append_pair("error_description", err.description());
+                },
+                _ => {}
+            }
         }
-        _ => {
+        url.into_string()
+    });
+    // Based on the kind of error, we may log the error, hide the description
```
djc: I think this comment could be expanded a bit. As I understand it, there are two basic resulting scenarios: (a) render the error template with the description set in the previous block (with appropriate status code), (b) redirect to the RP, with a specific URL also set in the previous block. The current comment makes it sound like there are three different things going on, which I find hard to trace/understand?

Also, maybe add a blank line before this whole block?

src/config.rs
```diff
@@ -89,8 +88,12 @@ impl Template {
             let (ref key, ref value) = param;
             builder = builder.insert_str(key, value);
         }
+        self.render_data(&builder.build())
```
djc: Now that we know this tweak is necessary, would it be feasible to put this hunk in the commit that introduces this code?

src/lib.rs
```diff
@@ -88,6 +89,33 @@ macro_rules! try_get_param {
 }
 
 
+/// Helper function for returning a result to the Relying Party.
```
djc: In this commit, I'd prefer to have `return_to_relier()` change in place (that is, don't change the order here), so it's easier to see what's going on here. You can of course move it around after that.

src/lib.rs
```diff
@@ -121,33 +121,17 @@ fn return_to_relier(app: &AppConfig, redirect_uri: &str, params: &[(&str, &str)]
 /// The `broker_handler!` macro calls this on error. We don't use a `From`
 /// implementation, because this way we get app and request context, and we
 /// don't necessarily have to pass the error on to Iron.
+///
+/// Based on the kind of error, we may log the error, hide the description
```
djc: Given the amount of restructuring here, this should probably be squashed into the previous commit that touched this code.

src/lib.rs
stephank: I'll do the latter. The reason it moved is specifically because of the reason mentioned in the comment: as soon as `client_id` and `redirect_uri` is verified, we should return errors to the RP.

src/lib.rs
stephank: Spoiler: the diff still covers the entire function ;)

stephank:
@djc Thanks for the review! I believe I've addresses all feedback.

src/email.rs
```diff
@@ -61,13 +61,16 @@ pub fn request(app: &AppConfig, email_addr: EmailAddress, client_id: &str, nonce
...
+        ("link", &href),
+    ];
     let email = EmailBuilder::new()
         .to(email_addr.to_string().as_str())
         .from((&*app.sender.address, &*app.sender.name))
-        .body(&format!("Enter your login code:\n\n{}\n\nOr click this link:\n\n{}",
-                       chars, href))
-        .subject(&format!("Code: {} - Finish logging in to {}", chars, client_id))
+        .text(&app.templates.email_text.render(params))
+        .subject(&format!("Finish logging in to {}", client_id))
```
djc: Why did you remove the code here? I thought it was rather useful. :)

djc:
Looking great, thanks!

src/email.rs
stephank: That snuck in, I guess. Think my reasoning when writing it was that I tried to de-emphasise the code in general, also with `<small>` tags in the HTML

Issue #32, opened 2016-10-17
Allow for setting configuration and signing keys via environment variables


callahad:
The only sane way to deploy configuration on a PaaS like Heroku is using env vars. See also: https://12factor.net/config

We need some way to pass configuration into the Broker without requiring a `config.json` and `key.pem` file on disk.

I'm particularly fond of a cascade of:

    App-Specific Env Vars > Common Env Vars > Config Files > Defaults

The demo rp is configured in this way: [settings.py](https://github.com/portier/demo-rp/blob/d693c2adcce2e4d2f53c1ea0ee9a33f636dd6ae9/settings.py), [README.rst#Configuration](https://github.com/portier/demo-rp/blob/d693c2adcce2e4d2f53c1ea0ee9a33f636dd6ae9/README.rst#configuration)

djc:
I read the 12factor thing, but I can't believe that the best solution for this problem space is to have 12 or 15 different environment variables to convey each of these values separately. And I presume the solution is also not having environment variables with JSON or some other way of having multiple actual values in there.

callahad:
For traditional servers or local testing, configuration files are still available, but for PaaS and container-based workflows, environment variables work extremely well and genuinely solve real problems -- they're the only thing that's feasible when you can't expect your local disk to be persistent.

It's also extremely important that we choose reasonable defaults, or try to infer as much as possible from the surrounding environment. For example, the [Demo RP](https://github.com/portier/demo-rp) has six available envvars, but I don't have to set any for local development, and I only have to set `DEMO_SECRET` for production deployments.

For the Broker, I suspect we'll only need to manually set `base_url`, `keys`, `sender`, the gmail `client_id`, and the gmail `secret`. Five variables isn't so bad.

It's also much simpler administratively: the variables persist across deployments, so we don't need to worry about sharing a private config file between us and carefully only checking it into a separate, private deployment branch of the repo. Instead, we just worry about the code, and the infrastructure worries about the configuration.

stephank:
Even `keys` is something we can generate for testing, with a big fat warning.

I haven't looked into it, but maybe you can sit inbetween `serde` and `serde_json`, implement this as a custom deserializer?

buro9:
Every setting should be settable via an ENV.

It's the only way to really manage process config and process secrets securely, and it's the preferred way in any container environment, i.e. Mesos + Marathon.

[...]

PR #42, opened 2016-10-18
Switch config from JSON to TOML

callahad:
In support of #32.

Currently based on #40; waiting for that to get merged before opening this up as a formal PR, but feel free to comment on my five commits at the end of this branch.


callahad:
This should be merge-able. r? @djc or @stephank

djc:
Should we get rid of `docopt` at this point? It doesn't seem to provide us with much value now. 

Why not put ip and port in a single addr config value? 

If the main advantage for TOML is that we can put comments in there, let's actually do so and get rid of the duplication between test file and README contents?

callahad:
> Should we get rid of docopt at this point? It doesn't seem to provide us with much value now. 

Probably.

> Why not put ip and port in a single addr config value? 

Cleaner semantics / granularity with regard to env vars. E.g., Heroku and other common PaaS's just provide `PORT` as an env var.

> If the main advantage for TOML is that we can put comments in there, let's actually do so and get rid of the duplication between test file and README contents?

On the way! :)

callahad:
Hitting the green button, given the approvals above

callahad:
Upon reflection, I still think docopt is useful, even if all it does is provide meaningful responses to `--help` and `--version`. Sure, we could manually do that, but docopt is already there, and it works.

Issue #45, opened 2016-10-21
Validate clients


stephank:
Currently, we don't validate `client_id` and `redirect_uri`, but I'm worried about the security implications of this.

OpenID Connect also has a spec on dynamic registration, which opens up the same hole, and discussed the implications here: http://openid.net/specs/openid-connect-registration-1_0.html#Impersonation

How do we validate clients while still keeping an open system?

stephank:
Random thought on this: maybe we can demand a `.well-known` entry on RPs as well. For example, we could say `client_id` is RP origin, and `redirect_uri` must be within that origin. Then, we request `{origin}/.well-known/portier-client`.

As for the contents, I'm not sure. Ideas:
 - We could do is use it as a challenge and implement OpenID Connect dynamic registration.
 - We could add crypto requirements to the nonce which we can validate using well-known, like a signature.
 - The well-known could also contain a display name. ("You are logging into Foobar (foobar.example.com)")

Downside of all of this is that we restrict RPs to those that are publicly reachable, which would be a real hassle for dev environments.

callahad:
Also, the burden of implementation moves us further out of reach / feasibility for smaller, hobbyist projects. Which I think is our sweet spot, or at least an under-served population.

callahad:
(But I'd completely support further exploration in this direction as an opt-in improvement for RPs.)

stephank:
Is anyone opposed to me implementing an optional feature to restrict clients in config, for those of us who want to run a private instance? (And if the option is missing, the broker is public, as it is now.)

That wouldn't resolve this ticket, but is more an extra.

onli:
@stephank: An option allowed_urls, which if set limit the allowed redirect_urls to those in that setting, would be fine imho

djc:
Can you explicitly state the threat you're worried about? It was my perception that the use of nonces in the RP side of the protocol mostly fix the problems around this stuff.

stephank:
I don't have a specific scenario, but it seemed like those two parameters are especially sensitive pieces of input, and we're handling them in a way that hasn't really been done before, as far as I know.

I hadn't considered the nonce, and that's definitely nice. Gives us a guarantee the login was actually initiated by the RP, as I understand.

djc:
Yeah, that's the goal of the nonce. Unless you can come up with a concrete threat/security problem, I think we should close this for now.

rfk:
> Random thought on this: maybe we can demand a .well-known entry on RPs as well.
[...]

PR #47, opened 2016-10-22
Optional whitelist of clients.

stephank:
As per discussion in #45.

I went with a bit more extended configuration, because it felt more natural. This matches the config we have for providers. The alternative, I think, would've been a flat list `redirect_uris`, but it'd be a toplevel field in the config.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -304,6 +304,11 @@ broker_handler!(AuthHandler, |app, req| {
     try!(valid_uri(client_id));
     try!(same_origin(client_id, redirect_uri));
     try!(only_origin(client_id));
+    if let Some(ref whitelist) = app.allowed_origins {
+        if !whitelist.contains(&client_id.to_string()) {
+            return Err(BrokerError::Input("the origin is not whitelisted".to_string()));
+        }
+    }
 
     // Per the OAuth2 spec, we may redirect to the RP once we have validated client_id and
     // redirect_uri. In our case, this means we make redirect_uri available to error handling.
--- a/README.rst
+++ b/README.rst
@@ -51,13 +51,14 @@ are:
 
 **[server] section:**
 
-============== ==================== ===========
-``config.ini`` Environment Variable Default
-============== ==================== ===========
-listen_ip      BROKER_IP            "127.0.0.1"
-listen_port    BROKER_PORT          3333
-public_url     BROKER_PUBLIC_URL    (none)
-============== ==================== ===========
+=============== ====================== =====================
+``config.ini``  Environment Variable   Default
+=============== ====================== =====================
+listen_ip       BROKER_IP              "127.0.0.1"
+listen_port     BROKER_PORT            3333
+public_url      BROKER_PUBLIC_URL      (none)
+allowed_origins BROKER_ALLOWED_ORIGINS (none) (unrestricted)
+=============== ====================== =====================
 
 **[crypto] section:**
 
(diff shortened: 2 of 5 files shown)


djc:
I'd like to understand the threat (as per #45) more. From preliminary review (i.e. if we merge this) I'd like to have this split up, so that the name thing, which seems to be an added nicety and non-essential for the whitelisting, is in a separate commit.

stephank:
I feel like this is useful even outside of the threat. We run a private installation, and I'd just like to limit it to our own apps.

Think I'll rid this of the `name`, part. We pretty much us `client_id` as display name, and the display name may just as well be the key in the config.

I'll wait for #48 to land before picking this up again.

stephank:
I've updated this with just a plain `redirect_uri` whitelist at `server.allowed_redirect_uris`.

callahad:
This seems reasonable as a stopgap, but I suspect we'll want something more robust down the road (e.g., whitelisting an entire origin, or even segregating whitelists based on API).

The OIDC spec also talks about pre-registration, API keys, and such. There's even an OIDC [Dynamic Client Registration](https://openid.net/specs/openid-connect-registration-1_0.html) spec, though iirc @rfk thought it to be a bit over-engineered.

At the same time, I presume you need this for Angry Bytes. If so, I'm happy to merge it as a temporary solution. Do you?

rfk:
> There's even an OIDC Dynamic Client Registration spec,
> though iirc @rfk thought it to be a bit over-engineered.

FWIW I think the data model of this spec is fine, but the burden of having to manage M x N client registrations is a bad fit for a truly decentralized systems with lots of reliers and lots of IdPs.  Using a discovery-based approach like suggested in [1] seems more likely to work out well from a long-term ecosystem perspective.

But that may or may not be a concern for you right now :-)

[1] https://github.com/portier/portier-broker/issues/45#issuecomment-255390270

stephank:
On the one hand, whitelisting by origin is nice, on the other hand, all OAuth2 servers I know of validate `redirect_uri` by exact value. So I feel like we should stick to validating exact values.

We can keep the client validation discussion in #45. If that results in some changes, then the whitelisting implementation can probably adapt. (E.g., if we start validating clients by origin, then we can whitelist origins, because there's some mechanism to get exact allowed `redirect_uri`s through discovery.)

src/lib.rs
```diff
@@ -298,10 +298,14 @@ broker_handler!(AuthHandler, |app, req| {
...
     // redirect_uri. In our case, this means we make redirect_uri available to error handling.
     let client_id = try_get_param!(params, "client_id");
-    let redirect_uri = try!(
-        Url::parse(try_get_param!(params, "redirect_uri"))
-            .map_err(|_| BrokerError::Input("redirect_uri is not a valid URL".to_string()))
-    );
+    let raw_redirect_uri = try_get_param!(params, "redirect_uri");
+    let redirect_uri = try!(Url::parse(raw_redirect_uri)
+            .map_err(|_| BrokerError::Input("redirect_uri is not a valid URL".to_string())));
+    if app.allowed_redirect_uris.len() > 0 {
```
callahad: nit: `!app.allowed_redirect_uris.is_empty()` instead of `app.allowed_redirect_uris.len() > 0`?

callahad: Or maybe this whole thing would be better as a an `Option<Vec<String>>`? Would this read more clearly?

```
if let Some(whitelist) = app.allowed_redirect_uris {
    if !whitelist.contains(&raw_redirect_uri.to_string()) {
        return Err(BrokerError::Input("the redirect_uri is not whitelisted".to_string()));
    }
}
```

stephank:
Rebased, made the field an `Option`, and it's now a whitelist of origins. (Since we patched `client_id`.)

PR #49, opened 2016-10-24
Add validations and fix redirect params.

stephank:
06402d7 is a small breaking change, but one I believe we need to do for better spec compliance.


callahad:
LGTM after rebasing. I'll also need to update the demo-rp to send `response_mode`

stephank:
@callahad Rebased!

PR #51, opened 2016-10-26
Address linter warnings from Clippy

callahad:
Nothing earth-shaking here, just some superfluous closures and borrows.


callahad:
(Let's sit on this until we clear the rest of the open PRs)

PR #72, opened 2016-10-28
Add validations for client_id and redirect_uri

callahad:


callahad:
@djc if you have a moment to review this, I'd greatly appreciate it. :)

src/validation.rs
```diff
@@ -0,0 +1,320 @@
...
+impl From<url::ParseError> for ValidationError {
+    fn from(err: url::ParseError) -> ValidationError {
+        ValidationError::Parse(err)
+    }
+}
+
+// -- Validation Functions --
+
+/// Test that a URI is valid and conforms to our expetations.
+pub fn valid_uri(raw_uri: &str) -> Option<ValidationError> {
```
djc: `Option<ValidationError>` doesn't feel like an idiomatic return type for this, I think `Result<(), ValidationError>` would be clearer even though it's a bit more typing.

src/lib.rs
```diff
@@ -39,7 +39,9 @@ pub mod email;
 pub mod oidc;
 pub mod store;
 pub mod store_cache;
+#[macro_use]
 pub mod validation;
+use validation::{valid_uri, only_origin, same_origin};
```
djc: I'd squash this commit into the previous one.

src/validation.rs
callahad: `Ok(())` felt weird, but you're probably right about being idiomatic. Rebasing, squashing, and adjusting to `Result<(), ValidationError>`

callahad:
Ready for re-review

callahad:
Whoops, didn't update the unit tests from `is_some` to `is_err`. Fixed, and actually ready for re-review now :)

Issue #38, opened 2016-10-17
Send reasonable security headers, including CSP, HSTS, and possibly HPKP


callahad:
References:
- https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#tab=Headers
- https://observatory.mozilla.org/

djc:
Currently the broker always runs behind a proxy, so I'm not sure if it makes sense to implement these in the broker itself or in the proxy?

callahad:
Since CSP tends to be rather deeply coupled to application structure and logic, I'm confident that one should at least be coming from the application.

buro9:
That and SRI https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

But it's a pain to maintain those.

callahad:
SRI is hard mode. We should aim for it, but not until the rest of our ducks are in a row. :)

PR #80, opened 2016-10-29
Revert "Require consent for oidc flow."

callahad:
This reverts commit 3d6d378be4a25c64e701e061809bee60d2af86fc.

The Google consent screen uses scary language:

screenshot from 2016-10-29 16-24-23

screenshot from 2016-10-29 16-24-36


stephank:
Well, that stinks, looks like we're running into this: http://stackoverflow.com/a/29267449/224129

We could revoke the access tokens, or implement implicit flow, but that'd require JS, because they don't support `response_mode=form_post`.

callahad:
Alternatively, if what we're concerned about is an invisible iframe initiating a login flow, we could just set `X-Frame-Options: Deny`, right?

stephank:
@callahad: We should probably set those headers regardless! :)

callahad:
Absolutely -- setting those headers is issue #38

PR #84, opened 2016-10-29
Revoke the access tokens we create.

stephank:
Alternative to: https://github.com/portier/portier-broker/pull/80

The scary warning may not go away until you remove access first, at: https://security.google.com/settings/security/permissions

From then on, all logins should show the same prompt, and as a plus we don't even show up in the connected apps list anymore.


stephank:
Closing in favour of https://github.com/portier/portier-broker/pull/86.

Issue #85, opened 2016-10-29
Set X-Frame-Options: Deny and/or Immediately Revoke Google Credentials


callahad:
Problem: After completing the Gmail flow, future logins do not require _any_ user interaction. This could be dangerous.

We tried using `?prompt=consent` (#76) to require reconfirmation, but it shows an alarming message to users (#80), so we can't do that.

Stack Overflow [suggests](http://stackoverflow.com/a/29267449/224129) that we could solve this by immediately revoking our access immediately after validating a token (#84).

It seems like we have three options:
1. Prompt on every login by immediately revoking our access token (#84)
2. Prompt on every login in the broker, before redirecting back to the RP.
3. Do not prompt on every login

The second option seems like over-engineering for now. So, do we want to prompt on every login or not? What attack are we actually trying to protect against?

callahad:
Option 1 requires a user click, and removes the broker from the user's Connected App dashboard, which could be a good thing.

Option 3 offers a streamlined user experience, at the potential cost of some security. If we're mainly concerned about sites trying to log people in by opening invisible iframes, maybe the `X-Frame-Options: DENY` header would be sufficient? It has [broad browser support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#Browser_compatibility)

stephank:
Maybe it's not just dangerous, but also too magic. If a user logs into site A with consent, then into site B without any consent at all, would that be scary?

callahad:
Possibly scary, but also possibly equal parts awesome.

callahad:
I like #84, but I'm *very slightly* leaning (~60%) toward waiting on that approach because I'm not confident in my answers to questions of its complexity:

- What happens if the revocation endpoint stalls. Do we block authentication until we succeed?
- Are there race conditions where a user might start logging into two websites, and the revocation somehow gets delivered in an order that breaks things?
- Do we automatically retry revocation if it fails?

stephank:
For what it's worth, we may only be seeing this 'fast' flow because we happen to use the authorization code OAuth flow. If Google implements the 'form_post' response mode, we could use the implict flow, which has less overhead, but likely also results in a prompt everytime.

onli:
There is something to be said for security by default, but if we don't know what we are protecting against we should opt for the easiest option. `X-Frame-Options: DENY` combined with doing nothing looks good to me.

stephank:
[...]

PR #86, opened 2016-10-30
Implement implicit flow for Google.

stephank:
Alternative to #84. See also the discussion in #85.

When using the implicit flow with Google, there is apparently never a prompt, not even on the first login. Perhaps that's because we already provide a `login_hint` as well.

We're also never listed in 'connected apps' in Google settings, because we don't have an access token. That's pretty good too. (However, if we want to send `prompt=consent` in the future, we _do_ get listed, and we _do_ get scary warnings the second login and beyond, like in #80. Maybe there are more options to control that.)

Google doesn't support `response_mode=form_post`, so we add a small page inbetween to capture the fragment parameters and POST them to us.

This removes a roundtrip on the server side, but adds a roundtrip on the client side. However, this also eliminates lingering access tokens, and maybe even preps us for supporting more OIDC providers in the future.

### DIFF
--- a/res/static/fragment_callback.js
+++ b/res/static/fragment_callback.js
@@ -1,0 +1,14 @@
+document.addEventListener('DOMContentLoaded', function() {
+  var form = document.getElementById('form');
+  var params = location.hash.slice(1);
+  var re = /([^&=]+)=([^&]*)/g;
+  var match;
+  while (match = re.exec(params)) {
+    var el = document.createElement('input');
+    el.type = 'hidden';
+    el.name = decodeURIComponent(match[1]);
+    el.value = decodeURIComponent(match[2]);
+    form.appendChild(el);
+  }
+  form.submit();
+});
--- a/src/handlers/oauth2.rs
+++ b/src/handlers/oauth2.rs
@@ -1,26 +1,49 @@
 use config::Config;
 use error::{BrokerResult, BrokerError};
 use iron::{IronResult, Plugin, Request, Response, Url};
+use iron::headers::ContentType;
+use iron::method::Method;
 use iron::middleware::Handler;
+use iron::modifiers;
+use iron::status;
 use oidc_bridge;
 use std::sync::Arc;
 use super::{RedirectUri, handle_error, return_to_relier};
-use urlencoded::UrlEncodedQuery;
+use urlencoded::UrlEncodedBody;
 
 
 /// Iron handler for OAuth callbacks
 ///
 /// After the user allows or denies the Authentication Request with the famous
 /// identity provider, they will be redirected back to the callback handler.
-/// Verify the callback data and return the resulting token or error.
+///
+/// For providers that don't support `response_type=form_post`, we capture the
+/// fragment parameters in JavaScript and emulate the POST request.
+///
+/// Once we have a POST request, we can verify the callback data and return the
+/// resulting token to the relying party, or error.
 broker_handler!(Callback, |app, req| {
-    let params = req.compute::<UrlEncodedQuery>()
-                    .map_err(|_| BrokerError::Input("no query string in GET request".to_string()))?;
+    match req.method {
+        Method::Get => {
+            Ok(Response::with((status::Ok,
+                               modifiers::Header(ContentType::html()),
+                               app.templates.fragment_callback.render(&[]))))
+        },
+        Method::Post => {
+            let params = try!(req.compute::<UrlEncodedBody>()
+                .map_err(|_| BrokerError::Input("no query string in POST data".to_string())));
 
-    let stored = app.store.get_session("oidc", &try_get_param!(params, "state"))?;
-    req.extensions.insert::<RedirectUri>(Url::parse(&stored["redirect"]).expect("unable to parse stored redirect uri"));
+            let stored = try!(app.store.get_session("oidc", &try_get_param!(params, "state")));
+            req.extensions.insert::<RedirectUri>(
+                Url::parse(&stored["redirect"]).expect("redirect_uri missing from session")
+            );
 
-    let code = try_get_param!(params, "code");
-    let (jwt, redirect_uri) = oidc_bridge::verify(app, &stored, code)?;
-    Ok(Response::with(return_to_relier(app, &redirect_uri, &[("id_token", &jwt)])))
+            let id_token = try_get_param!(params, "id_token");
+            let (jwt, redirect_uri) = try!(oidc_bridge::verify(app, &stored, id_token));
+            Ok(Response::with(return_to_relier(app, &redirect_uri, &[("id_token", &jwt)])))
+        },
+        _ => {
+            panic!("Unsupported method: {}", req.method)
+        }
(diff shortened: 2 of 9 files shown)


callahad:
Sitting on this until we get through the beta announcement. There's too much nuance in here for me to process while also prepping for launch :)

stephank:
Rebased on master, and now working with the CSP changes.

onli:
I had a look into this. 

If I understand https://openid.net/specs/openid-connect-core-1_0.html#Authentication (and some reading around that, but note the table) right, implicit flow is a good fit for us. It is after all what we use for our RP as well. I hesitated on `Client can be authenticated: no`, but the token gets authenticated, that is what counts for us. That the client can see the token does not really matter (signed + nonce), that not refresh tokens are possible is also no drawback for us. 

@callahad , you should still think through this, but for what it's worth, seems sound. And having less prompts is very nice. I'll see whether I find something wrong in the code now.

res/static/fragment_callback.js
```diff
@@ -0,0 +1,14 @@
+document.addEventListener('DOMContentLoaded', function() {
+  var form = document.getElementById('form');
+  var params = location.hash.slice(1);
+  var re = /([^&=]+)=([^&]*)/g;
```
onli: I'm not very happy we have to use a custom regex here. But I also have no better solution.

res/static/fragment_callback.js
```diff
@@ -0,0 +1,14 @@
+document.addEventListener('DOMContentLoaded', function() {
```
onli: If we use DOMContentLoaded, we run into the danger that that event might have already fired when we the JS is read. Sure, should only happen when including it async, but actually that is something we should do do gain a few performance pennys. Workaround is to check whether the event has already fired, as in https://github.com/onli/feedtragon/blob/161373ac3acd3738950e4bd4b0dacadd6eec4ccf/public/main.js#L431. I think that should be added.

src/handlers/oauth2.rs
```diff
@@ -1,26 +1,47 @@
...
+                               app.templates.fragment_callback.render(&[]))))
+        },
+        Method::Post => {
+            let params = try!(req.compute::<UrlEncodedBody>()
+                .map_err(|_| BrokerError::Input("no query string in POST data".to_string())));
 
-    let stored = app.store.get_session("oidc", &try_get_param!(params, "state"))?;
-    req.extensions.insert::<RedirectUri>(Url::parse(&stored["redirect"]).expect("unable to parse stored redirect uri"));
+            let stored = try!(app.store.get_session("oidc", &try_get_param!(params, "state")));
+            req.extensions.insert::<RedirectUri>(Url::parse(&stored["redirect"]).unwrap());
```
onli: The unwrap here is preferable to the expect + error message?

src/main.rs
```diff
@@ -83,6 +83,7 @@ fn main() {
 
         // OpenID Connect relying party endpoints
         get "/callback" => broker::handlers::oauth2::Callback::new(&app),
+        post "/callback" => broker::handlers::oauth2::Callback::new(&app),
```
onli: I tried to find it in the spec, but I did not find a clear statement that the implicit flow has to use a POST here. I missed it, or that is just how Google does it?

src/oidc_bridge.rs
```diff
@@ -78,14 +75,16 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
...
+        "&response_type=id_token",
         "&scope=",
         &utf8_percent_encode("openid email", QUERY_ENCODE_SET).to_string(),
         "&redirect_uri=",
         &utf8_percent_encode(&format!("{}/callback", &app.public_url),
                              QUERY_ENCODE_SET).to_string(),
         "&state=",
         &utf8_percent_encode(&session, QUERY_ENCODE_SET).to_string(),
+        "&nonce=",
+        &utf8_percent_encode(&nonce, QUERY_ENCODE_SET).to_string(),
```
onli: This reminds me: The spec mentions in https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest specifically for the implicit flow that the nonce needs to be unguessable. I guess with the client having access to the token that gets more important.

But If I see https://github.com/portier/portier-broker/blob/b72bcd195ed1503512ce662a085624bf6a4cbcef/src/handlers/oidc.rs#L120 right, we are using the nonce we get in the /auth request directly from the RP (I absolutely might be wrong here!). But if that is the case, we can't actually trust that nonce. We would need to create a nonce specifically for our own OIDC flow with the IdP.

We are probably already doing that and I'm just missing it in the Rust code?

src/oidc_bridge.rs
```diff
@@ -151,42 +146,8 @@ pub fn verify(app: &Config, stored: &HashMap<String, String>, code: &str)
                                           domain, e.description()))
         })?;
     let descr = format!("{}'s discovery document", domain);
-    let token_url = try_get_json_field!(config_obj, "token_endpoint", as_str, descr);
     let jwks_url = try_get_json_field!(config_obj, "jwks_uri", as_str, descr);
 
-    // Create form data for the Token Request, where we exchange the code
```
onli: All that just falls away? Very nice.

tmpl/fragment_callback.mustache
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Portier &ndash; Forwarding...</title>
+    <script src="/static/fragment_callback.js"></script>
```
onli: If adding the event check like mentioned above this script tag could get the async attribute.

src/oidc_bridge.rs
callahad: Great catch re: the nonce; we should be generating our own for the Broker <-> Google side of the transaction.

res/static/fragment_callback.js
stephank: Good idea, but I think it should also work to just move the script tag below the form. Will add a separate commit for this that also tackles `forward.js`.

src/main.rs
stephank: The OIDC spec says that `id_token` is supported with response modes `fragment` and `form_post`. Google only supports `fragment`, hence the `fragment_callback` trick.

The `GET` route here is repurposed for the `fragment_callback` trick, which is where users land with the `id_token` in the fragment. That page then copies fragment parameters to a form that calls the `POST` route here. (It's essentially a polyfill for `form_post` when the provider only supports `fragment`.)

I'm carefully avoiding calling the `GET` route here with the `id_token`, because it's dangerous according to the spec. I think that has to do with leaking the token in the `Referer` header.

stephank:
I rebased, fixed the unwrap, added `defer` to the scripts, and now added code to generate a separate nonce instead of reusing the RP nonce.

res/static/fragment_callback.js
onli: Okay, sounds good!

PR #87, opened 2016-10-30
Set CSP and X-Frame-Options on all responses.

### LINKED ISSUE #85: Set X-Frame-Options: Deny and/or Immediately Revoke Google Credentials

stephank:
Fixes #85.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,7 @@ extern crate urlencoded;
 
 use emailaddress::EmailAddress;
 use iron::headers::{ContentType, Location};
-use iron::middleware::Handler;
+use iron::middleware::{Handler, BeforeMiddleware};
 use iron::modifiers;
 use iron::method::Method;
 use iron::prelude::*;
@@ -67,7 +67,6 @@ macro_rules! broker_handler {
         }
         impl Handler for $name {
             fn handle(&self, req: &mut Request) -> IronResult<Response> {
-                info!("{} {}", req.method, req.url);
                 Self::handle_body(&self.app, req)
                     .or_else(|e| handle_error(&self.app, req, e))
             }
@@ -190,6 +189,16 @@ fn handle_error(app: &Config, req: &mut Request, err: BrokerError) -> IronResult
                                          ("error", "internal server error"),
                                      ]))))
         },
+    }
+}
+
+
+/// Middleware that logs each request.
+pub struct LogMiddleware;
+impl BeforeMiddleware for LogMiddleware {
+    fn before(&self, req: &mut Request) -> IronResult<()> {
+        info!("{} {}", req.method, req.url);
+        Ok(())
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 extern crate emailaddress;
 #[macro_use]
 extern crate log;
+#[macro_use]
 extern crate hyper;
 extern crate iron;
 extern crate lettre;
@@ -13,8 +14,8 @@ extern crate url;
 extern crate urlencoded;
 
 use emailaddress::EmailAddress;
-use iron::headers::{ContentType, Location};
-use iron::middleware::{Handler, BeforeMiddleware};
+use iron::headers::{ContentType, Location, StrictTransportSecurity};
+use iron::middleware::{Handler, BeforeMiddleware, AfterMiddleware};
 use iron::modifiers;
 use iron::method::Method;
 use iron::prelude::*;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -70,6 +75,14 @@ macro_rules! broker_handler {
...
                 Self::handle_body(&self.app, req)
                     .or_else(|e| handle_error(&self.app, req, e))
+                    .map(|mut res| {
+                        set_default_headers(&mut res);
+                        res
+                    })
+                    .map_err(|mut err| {
+                        set_default_headers(&mut err.response);
+                        err
+                    })
```
callahad: Does Iron have a notion of middleware that we could register, instead of manually doing this on each handler?

src/lib.rs
```diff
@@ -194,6 +207,22 @@ fn handle_error(app: &Config, req: &mut Request, err: BrokerError) -> IronResult
...
 
 
+/// Invoked after handlers for all responses, to set common headers.
+fn set_default_headers(res: &mut Response) {
+    // Specify a tight content security policy. We need to be able to POST
+    // redirect anywhere, and run our own inline scripts.
+    let csp = vec![
+        "sandbox allow-scripts allow-forms",
+        "default-src 'none'",
+        "script-src 'unsafe-inline'",
```
callahad: We should add a bug to further lock down script-src. Which would mean a followup PR that moves our JavaScript out into its own files.

src/lib.rs
```diff
@@ -194,6 +207,22 @@ fn handle_error(app: &Config, req: &mut Request, err: BrokerError) -> IronResult
 }
 
 
+/// Invoked after handlers for all responses, to set common headers.
+fn set_default_headers(res: &mut Response) {
+    // Specify a tight content security policy. We need to be able to POST
+    // redirect anywhere, and run our own inline scripts.
+    let csp = vec![
+        "sandbox allow-scripts allow-forms",
```
callahad: What's your thinking behind this sandbox value?

src/lib.rs
```diff
@@ -194,6 +207,22 @@ fn handle_error(app: &Config, req: &mut Request, err: BrokerError) -> IronResult
...
+        "default-src 'none'",
+        "script-src 'unsafe-inline'",
+        "form-action *",
+    ].join("; ");
```
callahad: Do we also need a style-src?

src/lib.rs
stephank: It does. I'll look into moving both this and logging to that.

src/lib.rs
stephank: Basically, 'why not?' 😋

We need the two options to do the POST redirects.

src/lib.rs
stephank: Not currently, but once we add style, yes. And it depends on how we add style.

src/lib.rs
stephank: Definitely! This seems a bit scary.

stephank:
I've added the extra headers and moved stuff to middleware.

src/lib.rs
```diff
@@ -194,6 +200,48 @@ fn handle_error(app: &Config, req: &mut Request, err: BrokerError) -> IronResult
...
+    fn set_headers(&self, res: &mut Response) {
+        // Specify a tight content security policy. We need to be able to POST
+        // redirect anywhere, and run our own inline scripts.
+        let csp = vec![
+            "sandbox allow-scripts allow-forms",
+            "default-src 'none'",
+            "script-src 'unsafe-inline'",
+            "form-action *",
+        ].join("; ");
+        res.set_mut((modifiers::Header(StrictTransportSecurity::including_subdomains(31536000u64)),
```
callahad: Let's send a plain HSTS header; I'm not confident we're ready for `includeSubdomains` yet, since my understanding is that it would apply to *all* domains of `portier.io`, or wherever the broker is running. And then have year-long consequences for a botched roll-out.

This could be especially dangerous for folks running their own brokers inadvertently locking themselves out of other, non-HTTPS subdomains on an origin.

For us, neither `https://portier.io` nor `https://demo.portier.io` are available, both redirect to another location *with* HTTPS.

With that changed, I'm +1 on merging.

callahad:
Closing in preference to #89

PR #89, opened 2016-10-31
Implement middleware and security headers

### LINKED ISSUE #38: Send reasonable security headers, including CSP, HSTS, and possibly HPKP

### LINKED ISSUE #85: Set X-Frame-Options: Deny and/or Immediately Revoke Google Credentials

callahad:
Identical to #87, but with HSTS not set to include subdomains.

Fixes #85 

Fixes #38

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,7 @@ extern crate urlencoded;
 
 use emailaddress::EmailAddress;
 use iron::headers::{ContentType, Location};
-use iron::middleware::Handler;
+use iron::middleware::{Handler, BeforeMiddleware};
 use iron::modifiers;
 use iron::method::Method;
 use iron::prelude::*;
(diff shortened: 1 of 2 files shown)


callahad:
Force-merging, as this was reviewed in #87, and we want to announce tomorrow.

PR #93, opened 2016-10-31
Fix/documentation

skorokithakis:
I've made a few fixes to the documentation, to make it more readable.

Main questions I have now:

* What's `keytext`?
* Is redis mandatory? If yes, aww :( If no, why does the readme say the redis address must be set?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -109,7 +109,7 @@ values for most settings, but you must explicitly set:
 
 * ``server.public_url``: The server's public-facing URL.
 * ``crypto.keyfiles``: An array of paths to encryption keys, or
-* ``crypto.keytext``: The text of an encryption key in PEM format
+* ``crypto.keytext``: The text of an encryption key in PEM format (takes precedence over keyfiles).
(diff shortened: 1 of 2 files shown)


skorokithakis:
@stephank, @djc, can either of you clarify the above questions? Ideally, I'd like to get this merged before launch to help first-time users.

stephank:
@skorokithakis Where keyfiles contains filenames, is an alternative to specify PEM contents directly.

Redis is mandatory, yes.

Also, for the changes to the README tables, please align them in the rst source too. :)

skorokithakis:
Ah, oops, good catch. So we either need to set `keyfiles` OR `keytext`?

skorokithakis:
Ah, never mind, the readme does say *or*. Thanks.

stephank:
Good question! It looks like, right now, you can specify both too. `keyfiles` can be keys, and `keytext` can just be one right now. If you specify both, `keytext` will be treated as 'after' the `keyfiles`.

Whichever key is last in the internal list is used as our signing key, so if you specify a `keytext`, it'll always be that one.

PR #95, opened 2016-10-31
Seperate form login from primary message

onli:
The important message is not the form, but that an email was sent. This minor styling change tries to support that by making the form elements smaller and adding a styled hr, also html semantics.

### DIFF
--- a/tmpl/confirm_email.mustache
+++ b/tmpl/confirm_email.mustache
@@ -17,7 +17,7 @@
           <em>{{ client_id }}</em>
         </p>
       </main>
-      < hr />
+      <hr />
       <aside>
         <p>
           Alternatively, enter the code from the email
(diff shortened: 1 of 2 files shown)


tmpl/confirm_email.mustache
```diff
@@ -8,24 +8,29 @@
...
-      </p>
-      <form id="form" action="/confirm" method="get">
-        <input type="hidden" name="session" value="{{ session_id }}">
-        <div class="entry">
-          <input type="text" name="code" maxlength="6" autofocus>
-          <button type="submit">Login</button>
-        </div>
-      </form>
+      <main>
+        <h1 class="head">
```
stephank: The heading and separator are now oddly aligned. Could you make sure all elements align?

onli: I did not see that as issue. But testing it out it does look better aligned. https://github.com/portier/portier-broker/pull/95/commits/98cd9912e8aa1e7d8eb75388ef01560f962bb287 fixes that for me.

stephank:
I aligned the `<h1>` as well. Will merge once travis finishes.

PR #97, opened 2016-11-01
[WIP] Resolve MX records before trying to mail.

stephank:
**DO NOT MERGE.**

I'm opening this so we can discuss it. But I feel this is not quite right:

 - We're going to need to rely on external services for the kinds of checks we want to do. This'll block threads. How happy are we with this?
 - In this case, the resolver instance also needs to be behind a mutex, so we're also blocking *other* threads. What can we do about this?
 - After writing this, I read this comment on HN. Apparently, an MX record is not required: https://news.ycombinator.com/item?id=12842689

Are there server frameworks out there yet which support futures?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,11 @@
 name = "portier_broker"
 version = "0.1.0"
 dependencies = [
+ "c-ares-resolver 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


stephank:
This'll be awesome once we have Iron + Hyper on top of Tokio. I'd feel a lot more comfortable if we weren't blocking a thread doing additional checks. Leaving this open so I can update it later on.

stephank:
Closing this because it's not moving along. :)

Issue #62, opened 2016-10-26
Improve logging and metrics


callahad:
Right now we have almost no visibility into the ongoing health and failure rates of this service.

We should make heavy use of our logging facilities, consider setting up some kind of log aggregation and error reporting service, and get some sort of statsd-style metrics going.

PR #98, opened 2016-11-02
Middleware refactor and canonical origin enforcement

### LINKED ISSUE #83: Create BeforeMiddleware that redirects incoming requests to $BROKER_PUBLIC_URL
We need middleware that ensure that hitting any origin other than the configured public url should 307 redirect to the configured origin.

E.g., hitting `http://portier-broker.herokuapp.com/foo?bar` should send me to `https://broker.portier.io/foo?bar`

callahad:
This pull request:

- Extracts middleware out of `lib.rs` and into its own module hierarchy.
- Refactors and adds basic unit tests for our security headers middleware.
- Introduces a new middleware to redirect incoming requests to the broker's canonical origin, as derived from `public_url`, which fixes #83.

### DIFF
--- a/src/middleware/headers.rs
+++ b/src/middleware/headers.rs
@@ -10,8 +10,8 @@ header! { (XXSSProtection, "X-XSS-Protection") => [String] }
 header! { (XFrameOptions, "X-Frame-Options") => [String] }
 
 /// Middleware that sets common headers.
-pub struct DefaultHeadersMiddleware;
-impl DefaultHeadersMiddleware {
+pub struct SecurityHeaders;
+impl SecurityHeaders {
     fn set_headers(&self, res: &mut Response) {
         // Specify a tight content security policy. We need to be able to POST
         // redirect anywhere, and run our own inline scripts.
@@ -30,7 +30,7 @@ impl DefaultHeadersMiddleware {
                      modifiers::Header(XFrameOptions("DENY".to_string()))));
     }
 }
-impl AfterMiddleware for DefaultHeadersMiddleware {
+impl AfterMiddleware for SecurityHeaders {
     fn after(&self, _: &mut Request, mut res: Response) -> IronResult<Response> {
         self.set_headers(&mut res);
         Ok(res)
--- a/src/middleware/headers.rs
+++ b/src/middleware/headers.rs
@@ -9,34 +9,56 @@ header! { (XContentTypeOptions, "X-Content-Type-Options") => [String] }
 header! { (XXSSProtection, "X-XSS-Protection") => [String] }
 header! { (XFrameOptions, "X-Frame-Options") => [String] }
 
-/// Middleware that sets common headers.
+/// Middleware that enforces common security headers on all outgoing responses.
 pub struct SecurityHeaders;
-impl SecurityHeaders {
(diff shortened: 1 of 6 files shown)


src/middleware/headers.rs
```diff
@@ -9,34 +9,56 @@ header! { (XContentTypeOptions, "X-Content-Type-Options") => [String] }
...
+        "style-src 'self'",
+        "form-action *",
+    ].join("; ");
+
+    res.set_mut((modifiers::Header(StrictTransportSecurity::excluding_subdomains(31536000u64)),
+                 modifiers::Header(ContentSecurityPolicy(csp.clone())),
+                 modifiers::Header(XContentSecurityPolicy(csp)),
+                 modifiers::Header(XContentTypeOptions("nosniff".to_string())),
+                 modifiers::Header(XXSSProtection("1; mode=block".to_string())),
+                 modifiers::Header(XFrameOptions("DENY".to_string()))));
```
stephank: The commit message mentions a scheme check, but that seems to be missing here.

src/middleware/origin.rs
```diff
@@ -0,0 +1,81 @@
...
+           uri.set_host(Some(&format!("{}", host))).is_err() ||
+           uri.set_port(Some(*port)).is_err() {
+            return Err(IronError::new(OriginError::Mutation, StatusCode::InternalServerError));
+        }
+
+        // ...and redirect to it.
+        let status = StatusCode::TemporaryRedirect;
+        let location = modifiers::RedirectRaw(uri.to_string());
+
+        Err(IronError::new(OriginError::Mismatch, (status, location)))
```
stephank: Downside here is that an `Err` gets logged, right? I've avoided logging anything for input errors. I'd prefer an AroundMiddleware here, if possible.

src/middleware/headers.rs
callahad: To be fair, it mentions scheme checks as "*potential* room for improvement" ;)

I thought CSP required a secure connection, but I appear to be mistaken. That means that only the HSTS header is unused on plaintext connections, where it just gets silently ignored. Not sure that's worth "fixing," especially since the broker should never run in production without HTTPS.

src/middleware/origin.rs
callahad: It does, but I still think it makes sense for this to be one of the first things requests hit -- we don't want to go through other potential middleware in the future (auth / session / logging / transaction stuff?) only to find out that we should abort and respond with a redirect.

I'll look into finding a way to suppress logging errors for this.

callahad:
@stephank If you're not strongly opposed, I'd like to proceed with this as-is, and tackle error handling / logging separately as part of #62. Thoughts?

stephank:
@callahad I am and I am not. :)

On the one hand, it bothers me, because it's now the only thing that generates spurious errors for simply bad input cases.

One the other hand, you're typically behind a proxy, it's an edge case that shouldn't be possible in production installs.

Can we add a TODO comment? I feel like it's something that'll be forgotten otherwise, even with #62.

callahad:
Sounds like a reasonable compromise :)

callahad:
TODOs added, commit message reworded.

Issue #69, opened 2016-10-26
Analyze one-time code entropy. is it enough?


callahad:
Brian Warner (@warner) had some interesting thoughts on this domain in his [PyCon talk on Magic Wormhole](https://www.youtube.com/watch?v=oFrTqQw0_3c) around the 12 minute mark.

Right now we [choose 6 characters](https://github.com/portier/portier-broker/blob/a88f503702e085c868d5aaaa13598ae1f6ba3988/src/email.rs#L44-L45) from [an array of 48 candidates](https://github.com/portier/portier-broker/blob/a88f503702e085c868d5aaaa13598ae1f6ba3988/src/email.rs#L16-L27).

48^6 = 12,230,590,464 combinations, or around 33.5 bits of entropy.

If we want to be case insensitive, we'd drop to just 26 possibilities:

26^6 = 308,915,776 combinations, or around 28.5 bits of entropy.

Is this enough?

callahad:
The Magic Wormhole stuff doesn't directly apply, but could still be useful context.

Since we don't have any out-of-band interaction, an attacker could force us to generate as many one-time codes as they want, and keep guessing perpetually, but with enough entropy and rate limiting (#70), we should be able to tune this to a reasonable rate of expected failure.

callahad:
Okay, let's do this. I don't have an academic understanding of probability and statistics, so I may get some of this wrong. Confirmation would be appreciated.

Let's presume:

- 1×10<sup>5</sup> requests per second (our server would melt well before this).
- No rate limits (naïve schemes could likely be circumvented by a determined attacker, sophisticated schemes would not likely be maintainable at our scale).
- Only one guess allowed per confirmation code.

There are roughly 3×10<sup>7</sup> seconds in a year (365.25 × 24 × 60 × 60 = 31,557,600).

That means a determined attacker gets 3×10<sup>12</sup> guesses per year, or 3×10<sup>14</sup> per century.

We want the attacker to have no more than a 1% chance every century, so the space of possible confirmation codes must be 3×10<sup>16</sup>, which is 100 times larger than the number of guesses.

Now we just need to solve for <em>n</em> in 2<sup><em>n</em></sup> = 3×10<sup>16</sup>:

    $ python3 -c 'from math import log; print(log(3e16)/log(2))'
    54.735812018918956

So as long as we have 55 bits of entropy, we should be good. Provided I haven't made some significant error, here.

callahad:
[...]

PR #100, opened 2016-11-04
Feature insensitive codes

### LINKED ISSUE #35: Lengthen and make case-insensitive email confirmation codes
If we're going to allow people to manually transcribe, being case insensitive is a huge usability win.

We should increase the length of the codes to compensate.

onli:
This should fix #35 by using only lowercase chars for generating the otp, and transforming the entered code to lowercase as well.

We have to think about the potential issues for #69 – I addressed this partly in the second commit by allowing more letters I think are not really confusing. Also, whether the code should be presented uppercase or lowercase in the mail was not clear to me. Steam for example uses uppercase. But it's easy enough to change that later, also without modifying how the code is generated.

### DIFF
--- a/src/email.rs
+++ b/src/email.rs
@@ -21,9 +21,7 @@ use url::percent_encoding::{utf8_percent_encode, QUERY_ENCODE_SET};
 /// and 'O'.)
 const CODE_CHARS: &'static [char] = &[
     '2', '3', '4', '6', '7', '9', 'a', 'c', 'd', 'e', 'f', 'g', 'h', 'j', 'k',
-    'm', 'n', 'p', 'q', 'r', 't', 'v', 'w', 'x', 'y', 'z', 'A', 'C', 'E', 'F',
-    'G', 'H', 'J', 'K', 'L', 'M', 'N', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W',
-    'X', 'Y', 'Z',
+    'm', 'n', 'p', 'q', 'r', 't', 'v', 'w', 'x', 'y', 'z'
 ];
 
 
@@ -41,8 +39,8 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
 
     let session = session_id(&email_addr, client_id);
 
-    // Generate a 6-character one-time pad.
-    let chars: String = (0..6).map(|_| CODE_CHARS[rand::random::<usize>() % CODE_CHARS.len()]).collect();
+    // Generate a 8-character one-time pad.
+    let chars: String = (0..8).map(|_| CODE_CHARS[rand::random::<usize>() % CODE_CHARS.len()]).collect();
 
     // Store data for this request in Redis, to reference when user uses
     // the generated link.
@@ -93,7 +91,7 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
 pub fn verify(app: &Config, stored: &HashMap<String, String>, code: &str)
               -> BrokerResult<(String, String)> {
 
-    if code != &stored["code"] {
+    if code.to_lowercase() != (&stored["code"]).to_string() {
         return Err(BrokerError::Input("incorrect code".to_string()));
     }
 
--- a/src/email.rs
+++ b/src/email.rs
@@ -90,7 +90,7 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
 pub fn verify(app: &Config, stored: &HashMap<String, String>, code: &str)
               -> BrokerResult<(String, String)> {
 
-    if code.to_lowercase() != (&stored["code"]).to_string() {
+    if &code.to_lowercase() != &stored["code"] {
         return Err(BrokerError::Input("incorrect code".to_string()));
     }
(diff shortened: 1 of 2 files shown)


src/email.rs
```diff
@@ -93,7 +90,7 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
 pub fn verify(app: &Config, stored: &HashMap<String, String>, code: &str)
               -> BrokerResult<(String, String)> {
 
-    if code != &stored["code"] {
+    if code.to_lowercase() != (&stored["code"]).to_string() {
```
callahad: We should probably apply the same transformation to both sides, otherwise we break all outstanding codes that have an upper-case letter when we deploy this.

src/email.rs
```diff
@@ -41,8 +38,8 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
 
     let session = session_id(&email_addr, client_id);
 
-    // Generate a 6-character one-time pad.
-    let chars: String = (0..6).map(|_| CODE_CHARS[rand::random::<usize>() % CODE_CHARS.len()]).collect();
+    // Generate a 8-character one-time pad.
+    let chars: String = (0..8).map(|_| CODE_CHARS[rand::random::<usize>() % CODE_CHARS.len()]).collect();
```
callahad: I'd really like an actual calculation of entropy and a threat model to guide us, here. The OWASP page on [Insufficient Session-ID Length](https://www.owasp.org/index.php/Insufficient_Session-ID_Length) is a good example of a calculation that could give us reasonable assurance.

tmpl/confirm_email.mustache
```diff
@@ -26,7 +26,7 @@
         <form id="form" action="/confirm" method="get">
           <input type="hidden" name="session" value="{{ session_id }}">
           <div class="entry">
-            <input type="text" name="code" maxlength="6" autofocus>
+            <input type="text" name="code" maxlength="8" autofocus>
```
callahad: Should we allow a longer length here, and then trim whitespace on the backend? Might help people who copy-and-paste the code, but accidentally select an extra space on either side.

src/email.rs
stephank: The `to_string()` here creates a copy. Instead, if you take a reference to the `String` left-hand, that'll coerce into `&str`, and allow comparison:

`if &code.to_lowercase() != &stored["code"] { ...`

stephank:
Random thoughts:

 - The OWASP example is about session IDs, not one-time pads. We have actual session IDs, and those are SHA256. The one-time pad is entropy on top of that.

 - The OWASP example still partially applies, as an example of brute-force time vs. session ID lifespan. Their calculation, however, accounts for session IDs that live for weeks, whereas both our session and one-time pad live for only 15 minutes.

 - The OWASP example doesn't account for rate-limiting. We can significantly reduce the rate of brute forcing with proper limiting.

 - More comparable to our one-time pad is TOTP, which is 10^6, or ~19.9 bits, with a window of 30s. If we simply extrapolate that to 15m, we need ~24.8 bits of entropy.

 - Some attacker scenario's, in order from good to bad:

   * Attacker knows absolutely nothing. He has to guess from 2^256 + 31^8 possibilities.
   * Attacker knows some of the SHA256 session ID input. That leaves him to guess the salt we add to the SHA function, so if I'm not mistaken: 256^16 + 31^8.
   * Attacker manages to get his hands on a valid session ID, 31^8 guesses remain.
   * Attacker has server access or intercepts a mail, game over.

onli:
We can move to 10 letters, moving us somewhere to 50 bit. It would not hurt us much, since those codes are only fallback.

However, that might be completely unnecessary if we are able to rate-limit how often one attacker can try to guess a code. I think we are able to do that. We don't have to go into limiting one how to limit one single attacker (and then trying to fight when the attack gets distributed), we can limit how often we verify a code for a single email address.

Also, the number of requests needed already with 8 letters  should be a lot more than our server could handle in 15 minutes. Our TTL really changes a lot already.

stephank:
I like that idea, simply allowing just 10 tries per session.

Even without rate limiting, I believe that if we can match TOTP, we're fine, right? And if the ~24.8 bits of entropy for 15 minutes calculation was correct, then we've always had more so far. Heck, we can have that with a pad of 8 **digits**. :)

src/email.rs
onli: Thanks! I fought with rust there.

stephank: I agree that it'd clutter the code, for fairly small breakage during beta. I'd be more worried if we were 1.0. :)

onli:
If we are okay to postpone the entropy question, then it is ready for review. max-size is now 12, withespace gets trimmed, and I removed the unnecessary to_string().

stephank:
One thing I missed is that our `/confirm` route currently allows guessing the session ID separately from the OTP. This means it's not just a simple addition 2^256 + 31^8, but two separate steps.

It's perhaps something we cannot change. If the user typos the code, we currently return to the RP with an error, but we probably should be rendering the form again. That means the two are always going to be separate steps.

callahad:
@stephank updated the last few bits of this in #103, closing :)

PR #103, opened 2016-11-06
Implement OTP using 12 chars of z-base-32.

stephank:
Implementation of 12 character codes from z-base-32, as discussed in #69.

This also ignores case and any other input outside the character set.

### DIFF
--- a/src/email_bridge.rs
+++ b/src/email_bridge.rs
@@ -13,18 +13,8 @@ use std::iter::Iterator;
 use url::percent_encoding::{utf8_percent_encode, QUERY_ENCODE_SET};
 
 
-/// Characters eligible for inclusion in the email loop one-time pad.
-///
-/// Currently includes all numbers, lower- and upper-case ASCII letters,
-/// except those that could potentially cause confusion when reading back.
-/// (That is, '1', '5', '8', '0', 'b', 'i', 'l', 'o', 's', 'u', 'B', 'D', 'I'
-/// and 'O'.)
-const CODE_CHARS: &'static [char] = &[
-    '2', '3', '4', '6', '7', '9', 'a', 'c', 'd', 'e', 'f', 'g', 'h', 'j', 'k',
-    'm', 'n', 'p', 'q', 'r', 't', 'v', 'w', 'x', 'y', 'z', 'A', 'C', 'E', 'F',
-    'G', 'H', 'J', 'K', 'L', 'M', 'N', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W',
-    'X', 'Y', 'Z',
-];
+/// The z-base-32 character set, from which we select characters for the one-time pad.
+const CODE_CHARS: &'static [u8] = b"13456789abcdefghijkmnopqrstuwxyz";
 
 
 /// Helper method to provide authentication through an email loop.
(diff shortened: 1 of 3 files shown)


src/email_bridge.rs
```diff
@@ -41,8 +32,14 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
...
-    // Generate a 6-character one-time pad.
-    let chars: String = (0..6).map(|_| CODE_CHARS[rand::random::<usize>() % CODE_CHARS.len()]).collect();
+    // Generate a 12-character one-time pad. For display, we split it in two
+    // groups of 6. We can use the unsafe conversion to str, because we stick
+    // to ASCII.
+    let chars_raw: Vec<u8> = (0..8).map(|_| {
+        CODE_CHARS[rand::random::<usize>() % CODE_CHARS.len()]
+    }).collect();
+    let chars = unsafe { from_utf8_unchecked(&chars_raw) };
+    let chars_fmt = [&chars[0..4], &chars[4..8]].join(" ");
```
callahad: This is generating 8-character codes :)

I benchmarked `from_utf8` and `from_utf8_unchecked` on my four year old ThinkPad, and the safe version is only 12 nanoseconds slower than the unsafe version. I'd rather pay the 12 nanosecond penalty and be able to claim "no unsafe code in Portier." This also protects us if, somehow, something outside of ASCII like a zero-width space somehow sneaks into `CODE_CHARS`.

src/email_bridge.rs
```diff
@@ -93,7 +90,8 @@ pub fn request(app: &Config, email_addr: EmailAddress, client_id: &str, nonce: &
 pub fn verify(app: &Config, stored: &HashMap<String, String>, code: &str)
               -> BrokerResult<(String, String)> {
 
-    if code != &stored["code"] {
+    let trimmed = code.to_lowercase().replace(|c| !CODE_CHARS.contains(&(c as u8)), "");
```
callahad: Should we be stripping out unrecognized characters? If we ever remove a character from `CODE_CHARS`, this has the side effect of invalidating unused codes generated by the previous version of the broker.

Not a problem for real people until we hit 1.0, but still... I'm at worst a -0 on this, but if an obvious, easy way to only strip whitespace occurs to you, let's do it.

src/email_bridge.rs
```diff
@@ -10,21 +10,12 @@ use super::{Config, create_jwt};
...
-/// (That is, '1', '5', '8', '0', 'b', 'i', 'l', 'o', 's', 'u', 'B', 'D', 'I'
-/// and 'O'.)
-const CODE_CHARS: &'static [char] = &[
-    '2', '3', '4', '6', '7', '9', 'a', 'c', 'd', 'e', 'f', 'g', 'h', 'j', 'k',
-    'm', 'n', 'p', 'q', 'r', 't', 'v', 'w', 'x', 'y', 'z', 'A', 'C', 'E', 'F',
-    'G', 'H', 'J', 'K', 'L', 'M', 'N', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W',
-    'X', 'Y', 'Z',
-];
+/// The z-base-32 character set, from which we select characters for the one-time pad.
+const CODE_CHARS: &'static [u8] = b"13456789abcdefghijkmnopqrstuwxyz";
```
callahad: Nice! 👍

stephank:
Pushed a new commit with fixes. It now just trims everything that matches `char.is_whitespace()`.

Issue #102, opened 2016-11-05
Disable autocomplete on form for confirmation codes


callahad:
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

Probably also worth adding the non-standard autocorrect="off" and autocapitalize="off" for iOS

stephank:
Fixed by #104

PR #104, opened 2016-11-14
Disable autocomplete for token input (#102)

onli:


tmpl/confirm_email.mustache
```diff
@@ -26,7 +26,7 @@
         <form id="form" action="/confirm" method="get">
           <input type="hidden" name="session" value="{{ session_id }}">
           <div class="entry">
-            <input type="text" name="code" maxlength="20" autofocus>
+            <input type="text" name="code" maxlength="20" autofocus autocomplete="nope" autocorrect="off" autocapitalize="off">
```
callahad: I believe we can use `autocomplete="off"` here, since we don't have any password forms. It's the attribute value Google uses for their own 2FA screen. Otherwise, r+ with that change.

PR #108, opened 2016-11-17
Remove origin check.

stephank:
It bugs me a bit, but I fear we have to remove the canonical origin check.

When hosted behind a reverse proxy, there's no indication of the actual scheme and port used to connect to the broker, except unreliable private headers (`X-Remote-*`). Often, the `Host` header may also be rewritten.

In this situation, the broker creates a redirect loop.


callahad:
There must be a reasonable way to preserve this... Maybe a config flag to trust (and require) [X-Forwarded-Proto/Port](https://devcenter.heroku.com/articles/http-routing#heroku-headers)?

callahad:
I tossed a little python app up at https://headerfoo.herokuapp.com/ which dumps all the request headers that a service running on Heroku sees...

stephank:
I'm worried that this is going to be a piece of code that is going to get more and more complicated.

Reading the implementation Express uses, `X-Forwarded-For` is a comma-separated list: https://github.com/jshttp/forwarded/blob/master/index.js

I've also seen proxies that instead send `X-Remote-Host` or `X-Remote-Addr`, and frameworks/apps that let you configure which header to look for.

There's a proposed standard [RFC 7239](https://tools.ietf.org/html/rfc7239) which implements an entirely different format, though I have not yet seen it in the wild.

I'm also worried that it's going to add a lot of configuration on the proxy end. If I'm not mistaken, both Nginx and Apache require explicit configuration to send these headers, and don't preserve the `Host` header by default.

But specifically the case where we are behind a reverse proxy, the most likely scenario and the scenario that'll require most configuration to make this work, we add no security benefit, because the proxy already enforces it.

stephank:
Part of the issue may eventually tie in with rate limiting. If we're going to limit on address, we will need to parse at least `X-Forwarded-For` to get the actual address too.

callahad:
I'll take another run at this later, but merging for now. On principle, I want to end up in a place where:

1. It is possible to securely host this on Heroku.
2. It is impossible to accidentally communicate with the broker over HTTP, and have it work.

Maybe reducing the scope to just enforcing https instead of the full origin would be a more reasonable path forward.

stephank:
Oh, didn't realise the reason for this was that the Heroku broker is accessible over plain HTTP.

It now bugs me more that Heroku doesn't have a simple switch to do this. :/

Issue #70, opened 2016-10-26
Rate-limit login attempts


callahad:
Dovetails with #69 -- how often should we allow a single email address + rp combo to request auth codes?

stephank:
So, I think we should:

 - Rate limit the number of sessions created per email.
 - Rate limit the number of code guesses per email (spanning all sessions).

On top of that:

 - Hard limit the maximum code guesses per session, after which we destroy the session.

And also:

 - NOT rate limit on IP. We should always be behind a proxy, which can do that better.

Does that sound right?

Some rate limiting implementation options:

 - Simple `INCR, EXPIRE, GET`. The nice thing here is that an attacker that proceeds past the limit will only extend his ban, without any extra actions going through. Downside is, the expiry should probably be short to not annoy regular users who trigger it by accident.

 - Sliding window using sorted sets, as per https://engineering.classdojo.com/blog/2015/02/06/rolling-rate-limiter/. This stores each action time as the score element of a sorted set, trims the set, then counts the elements. The implementation there may have a flaw of allowing an attacker to fill up memory by infinitely expanding the set. While this can be fixed by trimming by rank instead of score, it may still be troublesome if the maximum count is large.

 - Sliding window using hashes, as per https://gist.github.com/chriso/54dd46b03155fcf555adccea822193da. This splits up the window in buckets, where each hash key is a bucket with a count. The timespan of a single bucket determines granularity, but that means it'll also work well with long time windows.

Something to consider is to add several overlapping limits to a single type of action, e.g.: limit sessions per email to 3 times per 10 seconds, and 1000 per day.

I'm leaning towards the last implementation option, because it seems to handle all sizes of time windows well.

Thoughts and other options?

onli:
> NOT rate limit on IP. We should always be behind a proxy, which can do that better.

We can't guarantee that for every broker out there. If we see an easy way to add IP rate limiting on our side we should include it, especially if it is crucial for our security model. 

http://redis.io/commands/incr mentions rate limiting as an example for the incr command. I think that is what you mean with your first listed implementation option? It should be easy enough to adapt the code there to work with the email address instead of the IP. In this context, IP limiting might be worthwhile to implement directly as well (not as part of this rate limiting, but as an outer rate limit) to prevent the DOS-attack against one email address.

callahad:
> Downside is, the expiry should probably be short to not annoy regular users who trigger it by accident.

Or make it possible to trivial for an adversary to DDoS someone :)

[...]

PR #109, opened 2016-11-18
Add auth per email rate limit.

stephank:
First stab at an implementation of #70.

This is the third implementation option mentioned there, with I believe some small improvement. The details are explained here: https://github.com/portier/portier-broker/blob/d2374ae9ac2e0e3e306ce60fb5f666d8975ab406/src/store_limits.rs#L77-L86

### DIFF
--- a/config.toml.dist
+++ b/config.toml.dist
@@ -51,3 +51,10 @@ secret =
 discovery_url = "https://accounts.google.com/.well-known/openid-configuration"
 # Google OpenID Connect id_token issuer domain - Default: "accounts.google.com"
 issuer_domain = "accounts.google.com"
+
+
+[limits]
+# Limit the number of session of any kind.
+auth = ["per-second:3:1:1", "per-minute:20:60:5"]
+# Limit the number of email loop authentications.
+auth_email = ["per-second:1:1:1", "per-minute:10:60:5"]
--- a/src/store_limits.rs
+++ b/src/store_limits.rs
@@ -1,0 +1,118 @@
+use std::error::Error;
+use std::str::FromStr;
+use std::time::{SystemTime, UNIX_EPOCH};
+use super::redis::{self, PipelineCommands};
+use super::error::BrokerResult;
+use super::store::Store;
+use super::serde::de::{Deserialize, Deserializer, Error as SerdeError};
+
+
+/// Represents a limit configuration.
+///
+/// Typically provided as a string `<name>:<max_count>:<timespan>:<granularity>`.
+#[derive(Clone,Debug)]
+pub struct LimitConfig {
+    /// A name for this config, appended to the Redis key.
+    pub name: String,
+    /// Maximum request count within the window before we refuse.
+    pub max_count: u64,
+    /// Timespan of the entire window, in seconds.
+    pub timespan: u64,
+    /// Timespan of a single bucket in the window, which defines the
+    /// time granularity with which we store request counts.
+    pub granularity: u64,
+}
+
+impl FromStr for LimitConfig {
+    type Err = String;
+    fn from_str(s: &str) -> Result<LimitConfig, String> {
+        let parts: Vec<&str> = s.split(":").collect();
+        if parts.len() != 4 {
+            return Err("Limit config must be 4 colon-separated parts".to_string());
+        }
+
+        let max_count = parts[1].parse::<u64>().map_err(|e| e.description().to_string())?;
+        let timespan = parts[2].parse::<u64>().map_err(|e| e.description().to_string())?;
+        let granularity = parts[3].parse::<u64>().map_err(|e| e.description().to_string())?;
+        if timespan % granularity != 0 {
+            return Err("Limit timespan must be a multiple of granularity".to_string());
+        }
+
+        Ok(LimitConfig {
+            name: parts[0].to_string(),
+            max_count: max_count,
+            timespan: timespan,
+            granularity: granularity,
+        })
+    }
+}
+
+impl Deserialize for LimitConfig {
+    fn deserialize<D: Deserializer>(deserializer: &mut D) -> Result<LimitConfig, D::Error> {
+        let s = String::deserialize(deserializer)?;
+        LimitConfig::from_str(&s).map_err(|msg| D::Error::custom(msg))
+    }
+}
+
+
+/// Represents a Redis key.
+pub enum LimitKey<'a> {
+    Auth { email: &'a str },
+    AuthEmail { email: &'a str },
+}
+
+impl<'a> LimitKey<'a> {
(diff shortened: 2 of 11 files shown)


callahad:
I'm still wrapping my head around this strategy. Questions that come to mind:

1. Is `handlers/oidc.rs` the right place to be testing limits, or do we only care about ratelimiting for code-based (email_bridge.rs) workflows?

2. Should there be a higher level interface to `incr_and_test` to avoid the handler having to generate and submit the current time?

3. Can we avoid passing a reference to the store (`app.store.limits.incr_and_test(&app.store, ...)` feels a bit strange to me -- if it's a method chained down from `app.store`, why do we still have to pass `&app.store` as an argument? Or if we really must pass the reference, why is `limits` a member of `app.store` instead of a standalone function?

stephank:
1. Good point. Rate limiting emails is definitely the most important, but shouldn't we also limit sessions in general? A non-browser client could spam `/auth` with a GMail address to flood our Redis. Should it be a separate limit?

2. I originally tried to implement limiting by IP, but then realised we don't have the real IP, because we're behind a proxy. The idea was to measure time once, and use that everywhere. I somehow picked up that `gettimeofday` and friends are slow, and I was unsure what Rust is actually doing, but I'm now thinking it's probably stupid to optimise that in this case. Will change it.

3. This matches the design of `StoreCache`. I'm now thinking these could just as well be regular functions, instead of methods. We don't even need the `StoreCache` and `StoreLimits` structs or `self`. If that's okay, I'll add that change to the PR.

stephank:
I've rebased it and tried to address all three points.

Regarding the first, I've split up the limit in an `auth` limit for all kinds, and an `auth_email` for emails specifically.

config.toml.dist
```diff
@@ -51,3 +51,10 @@ secret =
 discovery_url = "https://accounts.google.com/.well-known/openid-configuration"
 # Google OpenID Connect id_token issuer domain - Default: "accounts.google.com"
 issuer_domain = "accounts.google.com"
+
+
+[limits]
+# Limit the number of session of any kind.
+auth = ["per-second:3:1:1", "per-minute:20:60:5"]
+# Limit the number of email loop authentications.
+auth_email = ["per-second:1:1:1", "per-minute:10:60:5"]
```
callahad: I wonder if the benefits of this degree of granularity are worth the complexity / cognitive overhead?

E.g., Could we automatically generate a reasonable granularity value? Do we need names for each limit, even though they might look like magic strings and confuse people? Should we deliberately turn them into magic strings? Do we need a vector, or could we get by with a single limit for each case?

Could we get by with a single stupid value like this, and would it be better?

```
[smtp]
per_user = "10/min"
```

I genuinely don't know what the best answer is to these would be. I do know that parsing the limits section hurts my head a little more than I'd like, but not enough to block merging of a patch that *does* work, as long as you're generally happy with this design.

src/store_limits.rs
```diff
@@ -0,0 +1,118 @@
...
+            .map(|i| (bucket + 1 + i) % num_buckets)
+            .collect();
+
+        let (counts,): (Vec<u64>,) = redis::pipe()
+            .atomic()
+            .hincr(&key_str, bucket, 1).ignore()
+            .hdel(&key_str, hdel_fields).ignore()
+            .expire_at(&key_str, (unixtime + config.timespan) as usize).ignore()
+            .hvals(&key_str)
+            .query(&store.client)?;
```
callahad: Would it make sense to try to merge all of the limit checks into a single Redis request? What's the overhead of hitting Redis four times for every email-based auth request? Eh, probably not worth worrying about. :)

callahad:
Wait, what about clock skew?

If the Broker is a few seconds behind Redis, we could end up setting an `EXPIREAT` in the past, and never actually enforcing a ratelimit.

If there's more than one Broker, they could disagree about which bucket is active, and could end up clearing buckets that should still be valid.

callahad:
Redis does have a `TIME` command that we could use to determine the current server time, but I wonder if we can somehow develop a scheme that doesn't rely on calculating timestamps, and which only uses the offset-based `EXPIRE` command?

config.toml.dist
onli: Just as a confirmation, when reading through that patch I though the same, that it would be nice to make that easier. I already have issues understanding the bucket system (never encountered that before), adding granularity on top of that in the config does not help.

callahad:
(I'm hesitant to rely on `TIME`, since network I/O could pause indefinitely between receiving the `TIME` response and issuing our subsequent query.)

callahad:
I found a few good articles on rate limiting strategies with Redis:

- Introduction to rate limiting with Redis: [Part 1](http://www.dr-josiah.com/2014/11/introduction-to-rate-limiting-with.html), [Part 2](http://www.dr-josiah.com/2014/11/introduction-to-rate-limiting-with_26.html)
- [Better rate limiting with Redis sorted sets](https://engineering.classdojo.com/blog/2015/02/06/rolling-rate-limiter/)

I really like that first article, and it seems to build up to a solution similar to the one in this pull request, and discusses all of the associated pitfalls and corner cases along the way. Ultimately, it ends up with a massive lua script that executes within the Redis server, which works, but seems like overkill. It also still requires two round-trips in order to first get the server's time, and then execute the custom script.

I'm starting to lean back towards a naive, single-bucket-per-email ratelimit, and only limiting the smtp provider. If we're defining our ratelimit in terms of requests per minute, I don't think we need to worry about smoothing across multiple buckets like we would if we were dealing with per-hour or per-day limits.

callahad:
Running a parallel branch at https://github.com/portier/portier-broker/compare/devil-in-the-details to see if another strategy is reasonably effective and sufficiently simple...

src/store_limits.rs
stephank: Four times, as in, four limit configs? The overhead of the individual queries in a single pipelined transaction is probably not bad. Not sure what the right tradeoff is otherwise.

config.toml.dist
stephank: I'm starting to think simpler is probably better too.

Calculating granularity is something I tried, but troublesome while we enforce `timespan % granularity == 0`. Something that could be worked around.

I think, as long as we use small windows, something like `10/min` per email with a single key could work just fine.

stephank:
I like the simpler strat. Was looking for `script load` in redis-rs, but apparently it'll lazy load scripts for you. Just have to create a `Script` and go. :)

onli:
Came here to propose the simpler strategy :) We should have an per-IP-rate-limiting around that, so one or a chosen small number of attackers can't block one email address. Can be on the proxy-level, or we just figure out how to reliably use X-Forwarded-For or another Proxy header iff behind a proxy, using the IP otherwise.

Issue #101, opened 2016-11-05
Compare known domains case-insensitively


callahad:
We don't delegate `@GMAIL.COM` to our Google handler. We should.

Edit: s/donations/domains. Thanks, autocorrect.

skorokithakis:
Do we have the automated tests working properly? They're a great way to add features like this one, because it's easy to develop them with TDD (rather than running through a whole flow to test). If so, I can spend a few minutes implementing this.

callahad:
Travis does run our unit tests, but I don't believe we have any around this functionality.

skorokithakis:
That's okay, I can write some for this (and develop against those). Thanks!

skorokithakis:
Has this been implemented already? I added a test but `assert_eq!(canonicalized("example@GMAIL.COM"), "example@gmail.com");` succeeds.

onli:
It has been done partly, in the rewrite of the canonicalization code in https://github.com/portier/portier-broker/blob/master/src/oidc_bridge.rs#L124. In fact, the only transformation if not gmail will be to lowercase the whole email address, and that is done always in the OIDC flow.

But the decision which flow will be taken is done in https://github.com/portier/portier-broker/blob/12c5eb25f29ef91be3a689feacc0e8f21d603f9d/src/handlers/oidc.rs#L108, and canonicalized is not called on the email at that time. And I'm not sure that would be wise. I think all we need to do is to lowercase `&email_addr.domain` – which sadly makes it less easy to write a good test for this.

PR #111, opened 2016-11-19
Lowercase the domain when deciding which flow to follow

skorokithakis:
Possible fix for #101.


callahad:
Doh. intuitively, this should work. In testing, it doesn't.

```
thread '<unnamed>' panicked at 'no entry found for key', ../src/libcore/option.rs:705
```

PR #112, opened 2016-11-19
Lowercase domains when looking up provider info

### LINKED ISSUE #101: Compare known domains case-insensitively

callahad:
Fixes #101
Completes #111


callahad:
(I've tested this end-to-end and it works as expected)

PR #113, opened 2016-11-22
Prevent whitespace from breaking template layout

callahad:
Note the space between the input widget and the submit button

---

screen shot 2016-11-22 at 17 40 21


onli:
We shouldn't need reviews for small stuff like that :/

Though it is actually a surprising you see that. I certainly have no white space between the two elements on the current broker, and I think there is no technical reason the html should cause one here?

callahad:
I think Firefox can just be pedantic about these things.

PR #114, opened 2016-11-22
WIP: Simplified rate-limiting

callahad:
Feedback on this approach?


stephank:
Squash from ae1e0ab down, and I'll happily merge. :)

PR #121, opened 2016-11-29
Add a metadata block and a `valid_from` to keys.

stephank:
This PR changes the output of config errors a bit, and we now fail when the keys could not be parsed. Example error:

```
$ cargo run -- ./config.toml
Failed to build configuration: failed to parse key 'private.pem': An error in the OpenSSL library: bad base64 decode
```

But the main purpose here is to add metadata to keys, with for now just a `valid_from`. This field is to facilitate key rotation, where we need to have a grace period during which we introduce the new key, but don't use it yet. This grace period allows clients with a cached keys document to keep functioning.


src/crypto.rs
```diff
@@ -12,66 +14,110 @@ use self::rustc_serialize::base64::{self, FromBase64, ToBase64};
...
+
+        let mut key = NamedKey::from_pkey(pkey)?;
+        key.valid_from = meta.valid_from;
+        Ok(key)
+    }
+
+    /// Parses metadata from a PEM file.
+    fn meta_from_pem(pem: &str) -> Result<PrivateKeyMetadata, String> {
+        toml::decode_str(
+            &pem.lines()
```
onli: There is no PEM library we can use to read an attribute in a pem code block? I did not find one, just checking that you checked as well.

src/crypto.rs
```diff
@@ -123,6 +169,14 @@ impl NamedKey {
             .insert("e", json_big_num(&e))
             .build()
     }
+
+    // Check if the key is valid at the given time.
+    pub fn is_valid_at(&self, time: &DateTime<UTC>) -> bool {
+        match self.valid_from {
+            Some(ref valid_from) => time >= valid_from,
+            None => true
```
onli: Do I understand that right that this will make the key be valid if valid_from is not set?

src/main.rs
```diff
@@ -50,23 +51,29 @@ struct Args {
 /// The `main()` method. Will loop forever to serve HTTP requests.
 fn main() {
     if let Err(err) = env_logger::init() {
-        panic!(format!("failed to initialize logger: {}", err.description()));
+        println!("Failed to initialize logger: {}", err.description());
+        exit(1);
```
onli: That is pretty minor, but I feel like others might prefer to have that exit code change in a separate PR. I think it does not matter much, it is a good improvement.

src/main.rs
```diff
@@ -50,23 +51,29 @@ struct Args {
...
     builder.update_from_common_env();
     builder.update_from_broker_env();
-    let app = Arc::new(builder.done().unwrap_or_else(|err| {
-        panic!(format!("failed to build configuration: {}", err.description()))
-    }));
+    let app = Arc::new(match builder.done() {
+        Ok(cfg) => cfg,
+        Err(err) => {
+            println!("Failed to build configuration: {}", err);
+            exit(2);
```
onli: But maybe that should be exit(3), as to signal the different type of error. Or do you think they are too close to be handled differently by other programs?

callahad:
I'd like to think more about the high-level design before delving into specific code review.

A few questions:

1. Does it make sense for this to be a property on the keys, instead of a configuration property?

2. Can we get by with an ordered list of keys, instead of explicitly annotating each with `valid_from`? E.g., always sign with the most recent key, but publish all configured keys?

stephank:
@callahad

1. That depends on what you have in mind?
2. That has always worked, and will continue working. `valid_from` is simply an extra limitation on when a key can be used, on top of 'only use the last key'.

My gripe with doing it in configuration is that it'd make the config section more complicated, and difficult the script. Here's what I'm currently doing in `cron.daily`:

```sh
#!/bin/sh
set -e
cd /srv/portier/cfg

# Generate the key.
if ! openssl genrsa -out next.pem 4096 2> /dev/null; then
    echo "Failed to generate new RSA key"
    exit 1
fi

# Append a metadata block to ensure the key is not used for the first hour.
tee -a next.pem > /dev/null << EOF
-----BEGIN PORTIER METADATA-----
valid_from = "$(date -u -d 'next hour' +'%Y-%m-%dT%H:%M:%SZ')"
-----END PORTIER METADATA-----
EOF

# Swap the keys.
mv current.pem previous.pem
mv next.pem current.pem

# Restart the container.
docker restart portier
```

src/crypto.rs
stephank: Yes! Keys that don't have the setting should simply work always.

src/crypto.rs
stephank: I didn't look, actually, but also think what this is doing is fairly specific. Most libs will only look for the key block.

src/main.rs
stephank: The difficulty in this is explaining the difference between 2 and 3, so I decided to keep them the same as just one class of 'config errors'.

stephank:
Rebased and fixed clippy errors.

callahad:
This still feels weird to me, as I've never run into this kind of metadata block in the wild. That said, I initially misunderstood the use case for this, and now that I understand it, I'm in favor of solving it *somehow*.

Other possible ways of solving this: 

- Encourage verifiers to (rate-limited!) always re-fetch cached JWKSets when encountering unrecognized key IDs, and call it somebody else's problem.
- In the broker configuration, specify keys as (path, Option<notBefore>) tuples, instead of embedding that information in the key file.
- Use two cron jobs: one that creates an `upcoming.pem` as the second entry in the key list, and another that promotes it to the head of the list an hour later.
- ... Something else? 

My gut feeling is partial to that third option, since it keeps the rotation logic and complexity completely separate from Portier, but it also has more moving parts. What are your thoughts on that approach?

If you believe this pull request represents the simplest, most standard, and most maintainable approach, then I'm all for it.

stephank:
You're likely right. If I simply modify the script to rotate 3 keys, then apply them in config shift by one (2, 3, 1), it should work as well. Will try that soon.

stephank:
Alright, the cron script now looks as follows, no longer depends on this PR, and seems to be working fine:

```bash
#!/bin/sh
set -e
cd /srv/portier/cfg

# Rotate 3 keys. The newest key is being introduced to clients before we use
# it. The oldest key is being phased out. The current key is the middle-aged
# key. Thus, we apply them in config as: 1. current, 2. newest, 3. oldest.

# The old 'current' key becomes the 'oldest' key.
if [ -f key1.pem ]; then
    mv key1.pem key3.pem
fi

# The old 'newest' key becomes the 'current' key.
if [ -f key2.pem ]; then
    mv key2.pem key1.pem
fi

# Generate a new 'newest' key.
if ! openssl genrsa -out key2.pem 4096 2> /dev/null; then
    echo "Failed to generate new RSA key"
    exit 1
fi

# Restart the container.
docker restart portier > /dev/null
```

And then I set the `BROKER_KEYFILES` using:

```bash
BROKER_KEYFILES=$(
    echo /srv/portier/cfg/*.pem |
        sed -e 's@/srv/portier@@g' |
        sort |
        sed -e 's@ @,@g'
)
```

(The extra `sed` is because `/srv/portier/cfg` is mounted as `/cfg` in the container.)

stephank:
I suppose we can close this now. Thanks for the scrutiny, guys! 😀👍

Issue #116, opened 2016-11-23
Figure out internationalization (i18n) / localization (l10n)


callahad:
We don't have a lot of text, but what we do have needs to be able to be localized.

onli:
https://github.com/portier/portier-broker/commit/cd67f14c4e8fc54fe9ca62a621f8edf3b28097fc implements an i18n scheme using gettext (using https://docs.rs/gettext/0.2.0/gettext/).

Some caveats:

 1. I used variable-like ids instead of the english default text as id, because I dislike having the english text in the sourcecode. I think that conflicts with how gettext is sued normally.
 * The auth handler, where the language detection is implemented currently, might be the wrong place.
 * There might be a better system than giving the text as variables to the mustache templates.
 * I was unable to add a config variable to our .toml scheme, I ran into an error there and just could not fix it. So instead this detects the language the browser requests via the Accept-Language header (which actually should be more difficult…). This would still need testing and probably modification if multiple languages are set in that header.
 * Those *.mo files are binary and maybe should not be in the repo, but then we need to setup Heroku to compile them
 * I could not test the email flow till the end, that still needs testing.

Ah, and most importantly: It seems to work fine :) If you think that approach can would be okay for us I'd like to open a PR.

callahad:
This is awesome! I don't have a lot of experience with i18n, so a few questions:

> I used variable-like ids instead of the english default text as id, because I dislike having the english text in the sourcecode. I think that conflicts with how gettext is used normally.

I think you're right that gettext is most commonly used with English literally in the source. I imagine this is useful since any changes to the English would automatically invalidate the translations. Without that, how else would we flag strings that need updates?

> I was unable to add a config variable to our .toml scheme, I ran into an error there and just could not fix it.

Yeah. The logic in the config stuff is good, but the source is really unwieldy. Prime target for a refactor after 1.0. Macros to the rescue? Or deeper Serde voodoo? There's got to be something. Updating https://github.com/softprops/envy to work with Serde 0.8 would let us remove [a few lines](https://github.com/portier/portier-broker/blob/master/src/config.rs#L421-L455).

> Those *.mo files are binary and maybe should not be in the repo, but then we need to setup Heroku to compile them

Would it make sense to compile them on the fly at server startup? What's a normal workflow there? Do folks really stop and manually recompile the .po files after every change?

stephank:
Really cool! Bummer that Mustache kind of breaks down here, and we can't do `{{ _ "Foo" }}` or similar.

[...]

PR #124, opened 2016-12-01
Support for i18n

onli:
See the discussion in #116.

This implements i18n via gettext, using the rust gettext module from https://github.com/justinas/gettext. A simple makefile is provided to help create the .mo files which the program reads. Translations are given as variables into the templates, but are strings in the sourcecode – that is not perfect, but https://github.com/nickel-org/rust-mustache does not support the better approach of calling a function from within the template.

I tested this till the point of sending the email out.

### DIFF
--- a/src/handlers/oidc.rs
+++ b/src/handlers/oidc.rs
@@ -3,7 +3,7 @@ use email_bridge;
 use emailaddress::EmailAddress;
 use error::{BrokerError, BrokerResult};
 use iron::{Handler, IronResult, Plugin, Request, Response, Url};
-use iron::headers::{ContentType, Location};
+use iron::headers::{ContentType, Location, AcceptLanguage};
 use iron::method::Method;
 use iron::modifiers;
 use iron::status;
@@ -14,7 +14,6 @@ use super::{RedirectUri, handle_error, json_response};
 use super::super::store_limits::addr_limiter;
 use urlencoded::{UrlEncodedBody, UrlEncodedQuery};
 use validation::{valid_uri, only_origin, same_origin};
-
 
 /// Iron handler to return the OpenID Discovery document.
 ///
@@ -115,21 +114,36 @@ broker_handler!(Auth, |app, req| {
     }
 
     if app.providers.contains_key(&email_addr.domain.to_lowercase()) {
-
         // OIDC authentication. Redirect to the identity provider.
         let auth_url = oidc_bridge::request(app, email_addr, client_id, nonce, &parsed_redirect_uri)?;
         Ok(Response::with((status::SeeOther, modifiers::Header(Location(auth_url.to_string())))))
 
     } else {
+        // Determine the language catalog to use.
+        let mut catalog = &app.i18n.catalogs[0].1;
+        if let Some(accept) = req.headers.get::<AcceptLanguage>() {
+            'outer: for accept_language in accept.iter() {
+                for &(ref lang_tag, ref lang_catalog) in &app.i18n.catalogs {
+                    if lang_tag.matches(&accept_language.item) {
+                        catalog = lang_catalog;
+                        break 'outer;
+                    }
+                }
+            }
+        }
 
         // Email loop authentication. Render a message and form.
-        let session_id = email_bridge::request(app, email_addr, client_id, nonce, &parsed_redirect_uri)?;
+        let session_id = email_bridge::request(app, email_addr, client_id, nonce, &parsed_redirect_uri, &catalog)?;
+
         Ok(Response::with((status::Ok,
                            modifiers::Header(ContentType::html()),
                            app.templates.confirm_email.render(&[
                                ("client_id", &client_id),
                                ("session_id", &session_id),
+                               ("title", catalog.gettext("Confirm your address")),
+                               ("explanation", catalog.gettext("We've sent you an email to confirm your address.")),
+                               ("use", catalog.gettext("Use the link in that email to login to")),
+                               ("alternate", catalog.gettext("Alternatively, enter the code from the email to continue in this browser tab:")),
                            ]))))
-
     }
 });
--- a/tmpl/email_html.mustache
+++ b/tmpl/email_html.mustache
@@ -3,21 +3,20 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>Portier &ndash; Finish logging in to {{ client_id }}</title>
+    <title>Portier &ndash; {{ title }} {{ client_id }}</title>
   </head>
   <body style="font: normal 1em/1.25em sans-serif">
     <div style="margin: 72px auto; max-width: 640px; text-align: center">
       <p style="margin: 24px; font:normal 1.25em sans-serif">
-        You received this email so that we may confirm your email address
-        and finish your login to: <em>{{ client_id }}</em>
+        {{ explanation }} <em>{{ client_id }}</em>
       </p>
       <p style="margin:24px">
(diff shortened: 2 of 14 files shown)


src/handlers/oidc.rs
```diff
@@ -115,21 +117,33 @@ broker_handler!(Auth, |app, req| {
...
         Ok(Response::with((status::SeeOther, modifiers::Header(Location(auth_url.to_string())))))
 
     } else {
+        let langtag_de: LanguageTag = "de".parse().unwrap();
+        let accept_language = req.headers.get::<AcceptLanguage>().expect("No accepted language set");
+        let lang_file = if langtag_de.matches(&accept_language[0].item) {
+            "lang/german.mo"
+        } else {
+            "lang/english.mo" // fallback, for easier catalog creation below
+        };
```
callahad: Is there a way to avoid hard-coding each language?

src/handlers/oidc.rs
```diff
@@ -115,21 +117,33 @@ broker_handler!(Auth, |app, req| {
...
 
     if app.providers.contains_key(&email_addr.domain.to_lowercase()) {
-
         // OIDC authentication. Redirect to the identity provider.
         let auth_url = oidc_bridge::request(app, email_addr, client_id, nonce, &parsed_redirect_uri)?;
         Ok(Response::with((status::SeeOther, modifiers::Header(Location(auth_url.to_string())))))
 
     } else {
+        let langtag_de: LanguageTag = "de".parse().unwrap();
+        let accept_language = req.headers.get::<AcceptLanguage>().expect("No accepted language set");
```
callahad: Fall back to English when no AcceptLanguage header is passed

src/handlers/oidc.rs
```diff
@@ -115,21 +117,33 @@ broker_handler!(Auth, |app, req| {
...
+        
         Ok(Response::with((status::Ok,
                            modifiers::Header(ContentType::html()),
                            app.templates.confirm_email.render(&[
                                ("client_id", &client_id),
                                ("session_id", &session_id),
+                               ("title", &catalog.gettext("Confirm your address")),
+                               ("explanation", catalog.gettext("We've sent you an email to confirm your address.")),
+                               ("use", catalog.gettext("Use the link in that email to login to")),
+                               ("alternate", catalog.gettext("Alternatively, enter the code from the email to continue in this browser tab:")),
```
callahad: I wish there was a way to distinguish between variables (`client_id` and `session_id`) and translated text in the templates. Is there a gettext helper for our templating system? Can we pass in the whole catalog and thus have something that looks like namespaces?

tmpl/email_html.mustache
```diff
@@ -3,21 +3,20 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>Portier &ndash; Finish logging in to {{ client_id }}</title>
+    <title>Portier &ndash; {{ title }} {{ client_id }}</title>
   </head>
   <body style="font: normal 1em/1.25em sans-serif">
     <div style="margin: 72px auto; max-width: 640px; text-align: center">
       <p style="margin: 24px; font:normal 1.25em sans-serif">
-        You received this email so that we may confirm your email address
-        and finish your login to: <em>{{ client_id }}</em>
+        {{ explanation }} <em>{{ client_id }}</em>
```
callahad: This would break for RTL languages... any idea of how to fix it? Not a blocker until we get our first contributed RTL translation. :)

src/handlers/oidc.rs
onli: Hm. We could probably try to select a translation file based on the langtag-language?

src/handlers/oidc.rs
onli: If we want to do things like that, I think we need to switch from mustache to handlebar or another template system. Is that something we want to tackle now?

tmpl/email_html.mustache
onli: The normal solution would be to give the variable to the gettext helper, with the translation placing the variable content where it belongs. Would be ugly here because of the html, but that can be ignore if necessary, but… frankly, I'm not sure whether https://docs.rs/gettext/0.2.0/gettext/ supports it, I don't see it clearly in the docs.

src/handlers/oidc.rs
onli: https://github.com/portier/portier-broker/pull/124/commits/52f83cd2b4dc5116951d1f9793e4c422e7602848 does that.

onli:
Short update: https://github.com/portier/portier-broker/pull/124/commits/52f83cd2b4dc5116951d1f9793e4c422e7602848 addresses the the comments I could address, namely the fallback to English if no language header is sent. I also tested that, it works. That commit also improves the code to actually go through all the language headers, properly in the order it is sorted on the client, and using the first found language it supports. That includes a basis to add more languages without bigger code changes, only the arrays linking language header to catalogue file would need an update if adding a new language (there might be a nicer language there, but this should work okay).

I asked @stephank to go through the code, as I again was fighting Rust trying to get this to work.

onli:
@callahad, @stephank ?

src/handlers/oidc.rs
```diff
@@ -115,21 +118,48 @@ broker_handler!(Auth, |app, req| {
...
+            for (index, langtag) in langtags.iter().enumerate() {
+                if langtag.matches(&accept_language.item) {
+                    lang_file = catalog_files[index];
+                    break 'outer;
+                }
+            }
+        }
+        
+        let f = File::open(lang_file).expect("could not open the catalog");
+        let catalog = Catalog::parse(f).expect("could not parse the catalog");
```
stephank: I'll repeat the comments from our private discussion on IRC here:

 - For consistency, use `to_string` instead of `to_owned`. (Converting from `&str` to `String`.) Not sure what the Rust preferred option is, but that's what we've used everywhere.

 - LanguageTag implements the `Clone` trait, which'd be a better option than two `langtag_en` variables.

 - I'd opt for keeping the language file names the same as the language tag, which could then simply be mapped to `lang/{}.mo`. We'd still probably want to hardcode the list of files in the code, but it'd be one less 'term' to deal with.

 - We should probably parse catalogs once on init, rather than on every request.

 - Feels like creating a default header value is a roundabout way of just initialising `lang_file` to the default language? When the for-loop falls through, it'd simply use that.

Here's a gist of how I'd write it: https://gist.github.com/stephank/a6c4647d16956335d2f48a62662362f8

onli: Thanks a lot @stephank. I pushed https://github.com/portier/portier-broker/pull/124/commits/7a43105d628d3f11f94b0347d0fb0360e23401a4, which is a mix of our two approaches. I'm very happy about the improvements you could show me.

To your comments:

 * I changed that to to_owned, but later on was able to remove that part completely
 * `Clone` is way better here, I searched that
 * Agreed and changed
 * We probably should, but I don't know how to implement that. Have a static(?) Vec<LanguageTag, Catalog> in config that is filled on init? That's too much rust for me. I instead adapted your code to only create one catalogue, the needed one.
 * Yes, that was pretty unnecessary. I did not understand how to use `if let Some(accept)`.

Could we merge this now? Or does someone want to move the catalogue parsing to init? I think that would be nice, but not something that should block this PR if no one wants to do it immediately. We can open an issue for  that.

stephank:
Pushed the following changes:

 - Squashed into one commit
 - Rebased on master
 - Translated email subject
 - Moved catalog parsing to config
 - Added a Dutch translation
 - Moved Makefile logic to `build.rs`
 - Updated `docker-build.sh` to package compiled catalogs

stephank:
Alright, clippy is now happy. I'll let you have one more look at the changes I've made, @onli, and if you think it's okay, you can hit that merge button! :)

onli:
Thanks, great!

PR #125, opened 2016-12-14
Workaround Clippy failures on current Nightly

callahad:


callahad:
Okay, we've got options:

- Disable Clippy on nightly for now
- Allow builds on nightly to soft-fail without blocking PRs
- Pin our rustc to nightly-2016-11-24 and clippy to 0.0.103 for now
- Any combination of the above

callahad:
The first two options don't get us any Clippy information at all, but we do want to know when the third option stops being necessary. So let's:

1. Allow builds on `nightly` to fail
2. Add a temporary `nightly-2016-11-24` build that is not allowed to fail

stephank:
Ref to clippy bug: https://github.com/Manishearth/rust-clippy/issues/1371

I like that last option. :)

Issue #115, opened 2016-11-23
Implement "dummy" SMTP provider


callahad:
For local testing, I'd love to just dump emails to my console instead of actually sending them.

onli:
I actually tested one for that, https://mailcatcher.me/, not integrated in the broker but as an external system. It did not work as the broker was dying with an error message about not having external access (or something similar) when trying to send the email to it. As that came from rusts mail internals I gave up on that point – but it still might be a good starting point, or an external solution we should just get to work.

callahad:
See: https://github.com/portier/portier.github.io/pull/27

callahad:
Woah, Mailcatcher looks cool.

stephank:
Or [MailHog](https://github.com/mailhog/MailHog)

onli:
So, I tried mailcatcher again. Configured like this:

```[smtp]
# Display name for confirmation emails - Default: "Portier"
from_name = "Portier"
# Sender address for confirmation emails - Default: (none)
from_address = "onli@Fallout"
# Outgoing mailserver address - Default: (none)
server = "smtp://localhost:1025"```
I also tried with 127.0.0.1. This is the error message:

```ERROR:iron::iron: Error handling:
Request {
    url: Url { generic_url: "http://localhost:3335/auth?scope=openid+email&response_type=id_token&nonce=d4a4f49c2ce04486a052c3c090d68cc8&response_mode=form_post&client_id=https%3A%2F%2Fportier-demo.herokuapp.com&login_hint=onli%40paskuda.biz&redirect_uri=https%3A%2F%2Fportier-demo.herokuapp.com%2Fverify" }
    method: Get
    remote_addr: V4(127.0.0.1:40070)
    local_addr: V4(127.0.0.1:3335)
}
Error was: Mail(Io(Error { repr: Custom(Custom { kind: Other, error: StringError("failed to lookup address information: Name or service not known") }) }))```

Doing some very profiessionial println!-debug, I now know that `let mut builder = SmtpTransportBuilder::new(app.smtp_server.as_str())?;` is the problematic line. Going even further down, I now know that `SmtpTransportBuilder::new` is the problematic part, as `app.smtp_server.as_str()` correctly prints as "smtp://localhost:1025". Which is the right value.

Don't know how to proceed. Someone of you familiar enough with lettre to know what is going wrong here? Or am I just missing something obvious?

callahad:
[...]

PR #131, opened 2017-01-25
Document dummy smtp option for dev (#115)

onli:


stephank:
I'm going to abuse this PR to fix the new clippy lints. :p

PR #132, opened 2017-01-25
Add cache headers.

### LINKED ISSUE #41: Send Cache-Control header in response
We should add the `Cache-Control` headers to our responses.
- Login routes: `no-cache, no-store`.
- Discovery document and other static resources: `public, max-age=86400` (1 day)
- Keys document should be `public`, with a configurable age. This depends on how often keys are rotated.
- `/.well-known`, probably `no-cache`?

stephank:
Fixes #41.

The approach here is to just set `Cache-Control: no-cache, no-store` for everything, except when overridden explicitly. The overrides are for static files, discovery and keys, each having a config var.

The discovery and keys resources don't set Last-Modified and Etag. I feel like it's more trouble than it's worth?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,6 +60,16 @@ public_url      BROKER_PUBLIC_URL      (none)
 allowed_origins BROKER_ALLOWED_ORIGINS (none (unrestricted))
 =============== ====================== =====================
 
+**[headers] section:**
+
+=============== ==================== ================
+``config.toml`` Environment Variable Default
+=============== ==================== ================
+static_ttl      BROKER_STATIC_TTL    604800 (1 week)
+discovery_ttl   BROKER_DISCOVERY_TTL 604800 (1 week)
+keys_ttl        BROKER_KEYS_TTL      86400 (1 day)
+=============== ==================== ================
+
 **[crypto] section:**
 
 =============== ==================== ================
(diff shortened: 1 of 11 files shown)


README.rst
```diff
@@ -60,6 +60,16 @@ public_url      BROKER_PUBLIC_URL      (none)
...
 =============== ====================== =====================
 
+**[headers] section:**
+
+=============== ==================== ================
+``config.toml`` Environment Variable Default
+=============== ==================== ================
+static_ttl      BROKER_STATIC_TTL    604800 (1 week)
+discovery_ttl   BROKER_DISCOVERY_TTL 604800 (1 week)
+keys_ttl        BROKER_KEYS_TTL      86400 (1 day)
```
onli: Having the key cacheable is nice, though also problematic if combined with the key rotation of https://github.com/portier/portier-broker/pull/121#issuecomment-274454789. If clients were to really respect the cache (which I doubt any client does, but let's assume) and the rotation happens during the time, validation should fail. 

At first I wanted to suggest to you to document that problematic, but since key rotation is external to the broker, I wouldn't know where would be a good place for that. So in the end I'm just leaving this as a comment here.

stephank:
I created #134, so we won't forget! :)

PR #138, opened 2017-02-04
Use serde_derive.

stephank:
Now that Rust 1.15 is out. 🎉

(We can't move to serde 0.9 yet, waiting on toml support.)


callahad:
FWIW, tracking bug for toml-rs supporting Serde 0.9 is https://github.com/alexcrichton/toml-rs/issues/135

PR #139, opened 2017-02-10
Upgrade to serde 0.9 and toml 0.3.

stephank:
- [x] Waiting on toml 0.3, then we're good to go! :)


stephank:
Alright, toml 0.3 is out, this should now be good to go!

PR #145, opened 2017-08-09
Hyper v0.11 and openssl crate v0.9

stephank:
This PR ditches Iron, uses plain Hyper 0.11, and upgrades the openssl crate to 0.9 for compatibility.

The entire request path up to calling bridges is async, and the OIDC bridge also does async HTTP requests to external systems. Mail and Redis are not async, but those are less critical I suppose.

The async changes don't really make the code any prettier. I've fought with finding a decent structure for things, but in general it's a whole lot of closures and moving things around.

Probably want to squash the commits at the end.


onli:
This looks good to me! @callahad, do you have time and motivation to do a detailed review? I will otherwise go over this next week, as far as I can, and greenlight then. Hope that works for you @stephank.

PR #150, opened 2017-10-27
Improve error messages and user feedback.

stephank:
This PR is on top of #149.

There are a bunch of changes to error handling here:

 - I've added `BrokerError::SessionExpired`, because I expect it to be a very common case, for which we can now give better feedback. The other classes of `BrokerError::Input` are all integration issues, and can use the generic template.
 - I've added `BrokerError::ProviderInput`, which covers some of the previous cases for `BrokerError::Provider` and `BrokerError::Input`. This is specifically for errors related to the request input (thus 400), not provider communication (still 503). All `BrokerError::Input` is now related to the RP, while all `BrokerError::Provider*` is related to the provider. Error messages reflect this.
 - Internal errors and rate limiting now no longer redirect back to the RP, and always display a message.
 - Internal errors are now logged with a reference number, which is also visible on the error page. This number is also z-base-32, so it can also be read on the phone, for example.
 - Error pages are now translated.

@onli Can you push a commit here with German translations for the new messages, if you can spare moment? :)

Example Dutch 'session expired' page:
screen shot 2017-10-27 at 15 24 55

Example Dutch 'internal error' page:
screen shot 2017-10-27 at 15 24 04

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,26 +2,26 @@
 name = "portier_broker"
 version = "0.2.0"
 dependencies = [
- "base64 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "base64 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "docopt 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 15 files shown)


onli:
Hi @stephank. Very nice. I added the german translation and approved the PR.

stephank:
Awesome, thanks! 😃

PR #151, opened 2017-12-04
Expose normalization as endpoint & upgrade dependencies

stephank:
This adds the `/normalize` endpoint, exposing the email normalization algorithm as a service. This is useful for languages/environments where it is difficult to implement the normalization. (It depends on good Unicode and WHATWG spec implementations.)

Portier returns a normalized email address in its tokens. Normalization is currently an issue anywhere the RP may want to compare user input to the result of Portier authentication. For example, an RP may store the email input before sending it off to Portier, or it may have a manually administered list of email accounts.


stephank:
This apparently breaks SMTP, because the default security mode used to be opportunistic. Will have to fix this.

stephank:
All good to go now. :)

stephank:
Made a small change to the way things work, so we can normalize multiple addresses in one call. I also rebased on master.

I've now built an integration using this in our app. When Portier finishes auth, and our app needs to lookup the user based on normalized email, we now lazily build a map of all users' normalized -> original email and cache it. This map is built with a single call `POST /normalize`. Then, our app can reverse the email normalization, and lookup the user in the database based on the original email.

This is just one of the ways to do things. An alternative method would've been to call `POST /normalize` when saving a user, so that the database always contains a normalized address. But either way, having a single call normalize multiple email addresses will support both of these approaches.

The API is fairly simple. Instead of `GET /normalize?email=xyz`, the client now does `POST /normalize` with a plain text body, one email per line. The response is also plain text, with matching lines of normalized emails. Bad input results in an empty line in the response.

PR #152, opened 2017-12-04
Fix coveralls

stephank:
Hopefully this'll work in one go. :)


stephank:
Alright, it's working again! Now using: https://crates.io/crates/cargo-travis

PR #172, opened 2018-06-24
prompt for login_hint if it wasn't presented

### LINKED ISSUE #166: Portier Broker could prompt for login_hint
The client is currently required to send:
- `login_hint`
- ~~`nonce`~~

As an example, Dex does not send these, so is currently not working with Portier.

~~I don't see that either of those should be explicitly required. It seems that `state` is used for security validation most times anyway, and that `portier-broker` could have another page that prompts the user for entering their email instead of every OIDC client needing to implement a way to get `login_hint` from the user.~~

I think `login_hint` could be omitted and Portier could prompt for it.

colemickens:
Fixes #166.

Screenshot:
screenshot

### DIFF
--- a/src/handlers/auth.rs
+++ b/src/handlers/auth.rs
@@ -122,14 +122,15 @@ pub fn auth(ctx_handle: &ContextHandle) -> HandlerResult {
 
     let login_hint = try_get_input_param!(params, "login_hint", "".to_string());
     if login_hint == "" {
+        let catalog = ctx.catalog();
         let data = mustache::MapBuilder::new()
             // TODO: catalog/localization?
             .insert_str("display_origin", &redirect_uri_)
-            .insert_str("title", "Login with email to")
+            .insert_str("title", catalog.gettext("Finish logging in to"))
             .insert_str("form_action", format!("{}/auth", &ctx.app.public_url))
             .insert_str("method", &ctx.method)
-            .insert_str("explanation", "Login with your email.")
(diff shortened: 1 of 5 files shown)


colemickens:
I've rebased this, and tests are passing.

src/handlers/auth.rs
```diff
@@ -109,12 +115,39 @@ pub fn auth(ctx_handle: &ContextHandle) -> HandlerResult {
...
-    let login_hint = try_get_input_param!(params, "login_hint");
     if try_get_input_param!(params, "response_type") != "id_token" {
         return Box::new(future::err(BrokerError::Input(
             "unsupported response_type, only id_token is supported".to_owned())));
     }
 
+    let login_hint = try_get_input_param!(params, "login_hint", "".to_string());
+    if login_hint == "" {
+        let data = mustache::MapBuilder::new()
+            // TODO: catalog/localization?
```
stephank: Is this something you'd like to try, or should we take a look at it?

colemickens: I went ahead and implemented it, following the pattern of other localized strings.

I'm not qualified to localize for other languages, though, unfortunately.

Tests are green as well. :slightly_smiling_face:

colemickens:
And the PR for the spec: https://github.com/portier/portier.github.io/pull/37

onli:
This can be merged, no?

onli:
Thank you! :)

PR #174, opened 2018-06-26
return verified_email as a bool (true) per oidc recommendations

### LINKED ISSUE #173: claim `email_verified` should be a boolean
See here: https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims

colemickens:
Fixes #173.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -225,7 +225,7 @@ pub fn create_jwt(app: &Config, email: &str, email_addr: &EmailAddress, aud: &st
     let payload = json!({
         "aud": aud,
         "email": email_addr.as_str(),
-        "email_verified": email_addr.as_str(),
+        "email_verified": true,
         "email_original": email,
         "exp": now + i64::from(app.token_ttl),
         "iat": now,


stephank:
We definitely want this, but it's another question of 'what will it break'. :/

There's currently not really an upgrade plan for the protocol, though technically we're not stable yet. We don't even announce these changes yet.

colemickens:
Understood. I'd be happy to move it behind a flag? Same with my other open PR? Let me know what you think.

onli:
I'm for merging it. We did not give stability guarantees yet, and the clients we know about we control and can change before the merge. Though the ruby client for example would not be affected by this. And we have to clear up the emails in the token (i.e. for normalization) anyway.

colemickens:
Would it help if I hosted an instance of Portier with this and the login_hint patch applied? And then I guess we could validate clients (do they each have tests, or example harness applications?)

stephank:
Well, all of the clients in the Portier organisation don't even use `email_verified`. It's more about what's out there we don't know of and can't test.

But you're right, @onli, though maybe we should send a short announcement to the mailinglist before deploying.

onli:
Thanks you for addressing this. I merged it, we already waited too long.

PR #177, opened 2019-01-25
Use the last occurence of @ in email normalization

jvasseur:
Update the broker implementation to match the spec changes in https://github.com/portier/portier.github.io/pull/38


onli:
Thank you!

PR #181, opened 2019-03-20
Remove IdP normalization

### LINKED ISSUE #165: Canonicalized email is returned as the 'sub' instead of the original email.
I tried to log in as `<first>.<last>@gmail.com`, but was instead authenticated as `<first><last>@gmail.com`.

I can't tell from @callahad's comment here if the "identical string" returned should be the original or the canonicalized version.

In my head, I can argue for both approaches. Just thought I'd ask and see if this is indeed intentional.

Thanks.

stephank:
Fixes #165

Suggest we keep protocol discussion in #165, and only discuss implementation details here.

This implementation uses only `email` (normalized) and `email_original` (original `login_hint`). So no `email_canonical`. The Broker does normalization on behalf of the RP if necessary, and there is no possibility for the IdP to change the email any more.

I will try deploy to schedule a deploy/transition on our private broker at $work in the meantime.

Spec changes in https://github.com/portier/portier.github.io/pull/39


onli:
> I will try deploy to schedule a deploy/transition on our private broker at $work in the meantime.

Did that work out fine? I see no issue apart from the discussed work RPs will ahve to do in their user management and will greenlight this.

stephank:
Didn't encounter any issues with the implementation itself, and it's been running for a while. 🙂

onli:
I merged this now. On my end I have to entangle this normalization a bit, and having this PR not in limbo will help :)

PR #184, opened 2019-07-15
Thwart email scanners that follow links (plus dep upgrades)

stephank:
We discovered some corporate environments have email scanners that follow links. This causes the scanner to invalidate the email auth session before the user can click the link.

This PR inserts a client-side redirect that translates the GET request to a POST request. This is similar to what OIDC did to transform `response_mode=fragment` to `form_post`, so that piece of code is now reused.

I've also done some dep updates, because I had trouble getting the latest stable Rust to build some dependencies.


stephank:
We've been running this on our own broker for a couple of weeks, without issues. 👌

PR #185, opened 2019-09-28
Fixes for the login form

stephank:
I apparently never created a PR for these fixes I had lying around locally. These are for the form displayed when no `login_hint` is provided.

@onli Maybe you can also add the two missing German translations?


onli:
@stephank I added the german translation :)

stephank:
Cool, thanks! 🙂

PR #186, opened 2019-09-30
Pin Heroku stack to heroku-16 for now.

stephank:
Should fix https://github.com/portier/portier.github.io/issues/40


greglearns:
Tested it by clicking on `https://heroku.com/deploy?template=https://github.com/portier/portier-broker/tree/fix/heroku`, and it works. LGTM.

Issue #183, opened 2019-04-10
Replace tokio-core with tokio


colemickens:
`tokio-core` is deprecated in favor of `tokio`. We should update Portier to use `tokio`.

I'm not totally sure of the state of things right now in async-rust though, maybe it'd be better to wait for more of async/await to settle and then move?

PR #189, opened 2019-11-24
Upgrade to Hyper 0.13, use async/await

### LINKED ISSUE #188: portier-broker requires openssl v 0.9.24 which is incompatible with openssl 1.1.1
I get this error when running cargo install.
```
error: failed to compile `portier_broker v0.2.0 (/opt/portier-broker)`, intermediate artifacts can be found at `/opt/portier-broker/target`

Caused by:
  failed to run custom build command for `openssl v0.9.24`

Caused by:
  process didn't exit successfully: `/opt/portier-broker/target/release/build/openssl-19a4f096748c3965/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'Unable to detect OpenSSL version', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.24/build.rs:16:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

stephank:
This is a pretty big deal for code quality, in my opinion.

We now have very little usage of `Box`, `Rc` and `Arc`, pretty much just in the outer layers `web.rs` and `main.rs`:
 - Our `Config` is still wrapped in an `Arc`. (We could `Box::leak` it? I'm not sure what to prefer.)
 - We need to pin-box our big request handling future type so we can provide a concrete type to Hyper's Service trait, and also satisfy its `Unpin` constraint.
 - Discovery uses `future::select`, which requires an `Unpin` future, so we have to pin-box that as well.

After this, the only remaining outdated dependency is Redis.

This should fix #188 and #183.


stephank:
I've done some more dep upgrades. Apparently `time` was completely reworked in 0.2, but it turns out we don't really need the dep.

I've deployed these changes to our company private broker, and will monitor that for a bit.

stephank:
Well, this is embarrassing. We apparently had cache expiry and session expiry times swapped this entire time. I thought I'd try it, because of the `time` crate changes.

Otherwise, testing has been good so far!

stephank:
This has been solid for us, so just going to merge it now. :)

PR #211, opened 2020-06-25
Additional email domain validation

stephank:
**EDIT: This description is old.** See: https://github.com/portier/portier-broker/pull/211#issuecomment-811704861

This is breaking and will require a 0.4 release.

This commit essentially reverts the email domain TLD validation, and moves it to a separate `DomainValidator` type allowing configuration (instead of the hardcoded list). This is also nice because `EmailAddress` is now again closer to our normalization spec.

`DomainValidator` also adds new stuff:

 - Validates against the [public suffix list](https://publicsuffix.org).
 - Adds simple allow/block-lists for email domains.

All of these lists can be read from files, and this has been extended to `allowed_origins`.


config.toml.dist
```diff
@@ -135,6 +129,55 @@ generate_rsa_command = ["openssl", "genrsa", "2048"]
...
+# The lists in this section all take literal string values, but also filenames
+# by prefixing a path with `@`. These files must contain one value per line,
+# but may contain empty lines or comments starting with `//` or `#`.
+#
+# Note that files are only loaded once during startup, and the broker must be
+# restarted after updating them.
+
+# List of (exact) website origins that are allowed to use this broker. If left
+# unset, the broker will allow any Relying Party to use it. (Note that an empty
+# list here has different meaning than leaving it unset.)
```
onli, line 144: Which one? Allowing no RP to pass?

config.toml.dist
```diff
@@ -135,6 +129,55 @@ generate_rsa_command = ["openssl", "genrsa", "2048"]
...
+################################################################
+# Access control
+
+# The lists in this section all take literal string values, but also filenames
+# by prefixing a path with `@`. These files must contain one value per line,
+# but may contain empty lines or comments starting with `//` or `#`.
+#
+# Note that files are only loaded once during startup, and the broker must be
+# restarted after updating them.
+
```
onli, line 141: It would be helpful to ass a newline here to separate the introduction setting from the next setting.

stephank:
I believe we had some discussions on Gitter about this in the past, but there are some issues with the lists approach:

 - Both the TLD list and the public suffix list appear to update often, the latter perhaps daily, which means we would need code to have the broker fetch it dynamically, plus configurability because our list code is generic.
 - The public suffix list is really just for the purpose of determining the public suffix from a domain, and not to validate a domain. There are real cases where using it for validation fails.

Instead, I scrapped both lists and implemented MX lookups here using trust-dns. I think this is a simpler, more foolproof approach.

I did keep the list code for `allowed_origins`, `allowed_domains` and `blocked_domains`, because it looks useful. Especially the private broker I manage for $work has a long list for `allowed_origins`, which would be great to have in a separate file.

jimdigriz:
Rust is not really something I can reliably read, but I could not see support for [Null MX](https://tools.ietf.org/html/rfc7505) records which has caught me out in the past.

stephank:
> Rust is not really something I can read, but I could not see support for [Null MX](https://tools.ietf.org/html/rfc7505) records which has caught me out in the past; though maybe you cater for it implicitly by the "must be an FQDN" which `.` may not be depending on the view of your validator.

@jimdigriz Ah, good catch! The `.` domain is a fqdn in trust-dns, but also easy to detect. We can definitely short-circuit.

The good news is that false-positives are not that bad, because we're only a first line of defence. The mail server would've eventually rejected it.

I've updated the PR and tested against `example.com`: https://github.com/portier/portier-broker/pull/211/commits/c85fc5e8f7e731c4b25262d219f4dfac422f5127#diff-fafaa9e37726cfb3918101190e0ff39c74895634a4cbd6b154cf4852e665b22cR158-R160

stephank:
Tempted to merge this if there are no more comments. :)

(The Mac CI failure is a fluke.)

PR #213, opened 2020-07-14
Add support for E2E testing of HTTP mailers

dstaley:
This PR adds support for testing the HTTP requests sent to HTTP mailers as part of the JavaScript E2E tests.

Based on the suggestions from @stephank, I added a new config option `BROKER_POSTMARK_API`, which is undocumented as it's only intended to be used during tests. The value of this environment variable is used as the endpoint where the postmark mailer sends HTTP requests. If not specified, it defaults to Postmark's API.

I've also added a new test utility, `postmarkTest`, that only runs the test when `TEST_MAILER` is equal to `postmark`.

The approached used in this PR lay the foundation for adding new HTTP-based email APIs such as Mailgun.

(Also, as a small tweak, I've added the `node` executable to the npm test script since Windows doesn't support specifying interpreters with the hashbang syntax.)

### DIFF
--- a/src/agents/mailer/postmark.rs
+++ b/src/agents/mailer/postmark.rs
@@ -41,10 +41,6 @@ impl Agent for PostmarkMailer {}
 
 impl Handler<SendMail> for PostmarkMailer {
     fn handle(&mut self, message: SendMail, cx: Context<Self, SendMail>) {
-        // If using the test token, print the email text body to stderr.
-        // This is here for our test suite, but maybe useful in general.
-        let is_test_request = self.token == "POSTMARK_API_TEST";
-
         let body = serde_json::to_vec(&json!({
             "From": &self.from,
             "To": message.to,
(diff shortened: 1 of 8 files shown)


dstaley:
@stephank maybe I'm misunderstanding your comments, but it sounds like you don't want mailer-specific tests? With this approach we can test specific attributes of the API request, such as the presence of the `from` field. This might not make too much sense currently, but one of my planned features with the Mailgun mailer was adding support for AMP for Email, which is specified via a specific request property.

stephank:
Ah, I wasn't aware that was your intention. In that case, sounds good!

Though now that we have a mock Postmark endpoint, I would like to get rid of the `is_test_request` path inside the Postmark agent. I think if you call `mailbox.pushMail()` from within the http-mailer request handler (in addition to `requests.push()`), that should make E2E tests work again.

Btw, I wasn't aware of amp for email, so curious: what can it do in Portier context?

dstaley:
Okay cool, I'll take a look at switching it to expect mails from the server instead of stdout on HTTP mailer tests!

The biggest improvement I see in AMP for Email for Portier is the ability to bypass the antiquated email rendering engines that most email clients use. As of now the only AMP consumers are Gmail, Outlook for Web, and mail.ru, but I anticipate usage growing over the next few years. One particular aspect of AMP for Email that Portier could utilize is the ability to dynamically determine if a login link is still valid. So, for example, AMP emails could disable the login button and replace it with a "Whoops, looks like this link has expired" text.

PR #222, opened 2020-11-06
Upgrade dependencies, clippy fixes, add caching to CI

stephank:
Some of our deps required Rust 1.42, plus we now use `matches!` ourselves, so the minimum version was bumped.

I've also re-enabled clippy and fixed some new errors that were generated.


onli:
Something specific to check here? The changes seem reasonable, we could merge it.

stephank:
Not really, guess I just feel bad if I push stuff through all on my own. 😅

PR #227, opened 2020-12-28
Bump redis crate to v0.17.0

dstaley:
`redis` v0.17.0 [added support for Redis 6 auth](https://github.com/mitsuhiko/redis-rs/blob/master/CHANGELOG.md#0170---2020-07-29), which many hosted Redis providers are now offering. The main feature of Redis 6 that portier is currently incompatible with is username/password auth.

This PR bumps the `redis` crate's version to the oldest version with support for Redis 6 auth. Versions v0.18 and v0.19 contain larger changes (including an upgrade to tokio v1), so I wanted to keep this update as small as possible since my Rust skills aren't that great.

Upgrading required a few changes:

1. `parse_redis_value_async` now expects a decoder as the first argument. In `redis`, this decoder comes from the `combine` crate, so that's been added as a dependency.
2. v0.17.0 also supports TLS connections, so the Rust compiler is ensuring that portier is handling that option. However, since adding those options would be a bit more complicated, I've simply returned an error stating that TLS connections aren't supported yet.
3. `pubsub.rs` was updated to authenticate with username/password when both are provided.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,16 +22,21 @@ dependencies = [
 ]
 
 [[package]]
-name = "ascii"
-version = "0.9.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eab1c04a571841102f5345a8fc0f6bb3d31c315dec879b5c6e42e40ce7ffa34e"
-
-[[package]]
 name = "ascii_utils"
 version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


stephank:
Looks great! Will tag a release in a bit.

Had to fight CI on this one, but it looks like the cache on macOS was corrupt, and there's no easy way to clear it. So I added a silly cache bust in ef7484a494946188348b2ccc3b2f4bb9fa9a026c. 🙃

PR #236, opened 2021-03-11
Preserve combine decoder instance in pubsub

### LINKED ISSUE #234: Redis connection issues with 0.3.4
I'm seeing some weird connectivity issues with Redis on 0.3.4. After a little while, the broker crashes with:

```
thread 'tokio-runtime-worker' panicked at 'Failed to read from Redis: parse error: Parse error at 1
Unexpected `117`
Unexpected `10`
', src/utils/redis/pubsub.rs:167:33
```

stephank:
Should fix #234.

Turns out a fresh combine `Decoder` for every `read()` no longer works, we need to preserve it. Appears to have fixed it locally on my Mac.

### DIFF
--- a/src/utils/redis/pubsub.rs
+++ b/src/utils/redis/pubsub.rs
@@ -13,18 +13,27 @@ use tokio::sync::{broadcast, mpsc, oneshot};
 #[cfg(unix)]
 use tokio::net::UnixStream;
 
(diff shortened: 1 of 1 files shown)


stephank:
Going to ignore the macOS failures. One is a fluke, one is another spurious hang. :/

PR #240, opened 2021-04-20
[RFC|WIP] deploy to GCP

jimdigriz:
**WARNING:** Work in Progress, do not merge.

Some client work I was doing including making Portier deployable to GCP so I took the opportunity to generalise the solution so that others should be able to use it too. The deployment is described in the documentation in my branch at https://github.com/jimdigriz/portier-broker/tree/gcp/contrib/gcp and includes a list of issues/limitations; unlikely to be solved by myself as most require a Rust coder to work on the actual broker.

Probably of most interest to the project is the [(non-awful) conditional build](https://stackoverflow.com/a/54245466) feature slipped into the `Dockerfile` that lets the user deploy a non-forked version of the project in their environment and pull in the data directory as an externally sourced tarball from an HTTP server.

As for the GCP deployment, it is the usual meat and potatoes you would expect...both the good and bad...

The project is far enough along to start seeking feedback from the maintainers and, if there is an appetite for it, what changes would be necessary for consideration of inclusion in the project.

### DIFF
--- a/contrib/gcp/README.md
+++ b/contrib/gcp/README.md
@@ -1,0 +1,153 @@
+Portier Deployment to Google Cloud Platform (GCP).
+
+The deployment fronts Portier with Redis running under [Cloud Run](https://cloud.google.com/run) with an [external cross-region global load balancer](https://cloud.google.com/load-balancing/docs/https#cross-region_load_balancing) providing [resilience to region failure](https://cloud.google.com/run/docs/multiple-regions) (though only a single region is used).
+
+Baseline costs are ~$80/month when idle:
+
+ * [Cloud Run](https://cloud.google.com/run/pricing#tables) (~$10/month)
+     * You may wish to pick a Tier 1 region for better pricing where possible
+ * [Redis](https://cloud.google.com/memorystore/docs/redis/pricing#instance_pricing) (~$50/month)
+     * Uses standard HA tier with 1GiB capacity
+ * [Load Balancer](https://cloud.google.com/vpc/network-pricing#lb) (~$20/month)
+     * Deployment fits within the first five forwarding rules pricing
+
+## Related Links
+
+ * [Deployment Manager](https://cloud.google.com/deployment-manager)
+ * [Jinja Template Documentation](https://jinja.palletsprojects.com/en/2.11.x/templates/)
+
+## Outstanding Issues
+
+ * only a single region as GCP's Redis HA deployment only operates within a single region and we have to use the same instance everywhere as otherwise the server 'client' could hit a different backend to the browser 'client'
+ * The [type providers](https://cloud.google.com/deployment-manager/docs/configuration/type-providers/advanced-configuration-options) used need massaging to allow updates
+ * Hard coding of the Redis port to `6379/tcp`
+     * workaround: `Reference [$(ref.portier-europe-west4-redis.port)], was not of type string but [NUMBER], cannot replace inside ["redis://:abcd@192.0.2.1:$($(ref.portier-europe-west4-redis.port))/0\n"]`
+     * GCP's Deployment Manager is mostly awful, so when GCP throws you lemons, it provides zero tools (or documentation) to make lemonade
+         * there is no cast operator available that works on `$(ref ...)`
+         * we cannot use [`outputs`](https://cloud.google.com/deployment-manager/docs/configuration/expose-information-outputs) as it uses pass by reference and creates the same problem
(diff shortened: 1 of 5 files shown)


contrib/gcp/README.md
```diff
@@ -0,0 +1,143 @@
...
+## Outstanding Issues
+
+ * The [type providers](https://cloud.google.com/deployment-manager/docs/configuration/type-providers/advanced-configuration-options) used need massaging to allow updates
+ * Hard coding of the Redis port to `6379/tcp`
+     * workaround: `Reference [$(ref.portier-europe-west4-redis.port)], was not of type string but [NUMBER], cannot replace inside ["redis://:abcd@192.0.2.1:$($(ref.portier-europe-west4-redis.port))/0\n"]`
+     * GCP's Deployment Manager is mostly awful, so when GCP throws you lemons, it provides zero tools (or documentation) to make lemonade
+         * there is no cast operator available that works on `$(ref ...)`
+         * we cannot use [`outputs`](https://cloud.google.com/deployment-manager/docs/configuration/expose-information-outputs) as it uses pass by reference and creates the same problem
+ * support for [in-transit encryption](https://cloud.google.com/memorystore/docs/redis/in-transit-encryption)
+     * [redis crate supports it](https://docs.rs/redis/0.20.0/redis/enum.ConnectionAddr.html#variant.TcpTls) though Portier's [`pubsub.rs`](../../src/utils/redis/pubsub.rs) explicitly does not
```
stephank, line 30: Ah, sorry about that. I'm not super happy with the pubsub code, but also haven't really worked with upstream to get us everything we need. Maybe newer versions already do what we want. (Can't say, don't remember off the top of my head what the exact pain points were.)

stephank (commented):
Thanks for working on this! This is super interesting as someone who works at an AWS + Terraform shop. 😃

I'm no expert in any of the gcloud stuff, so that looks good to me. Just want to flesh out the Docker part some more, because I consider it one of the primary installation methods. (Even though I haven't paid enough attention to it.)

Dockerfile
```diff
@@ -2,45 +2,62 @@
...
+  && apt-get -y --option=Dpkg::options::=--force-unsafe-io upgrade --no-install-recommends \
+  && apt-get -y --option=Dpkg::options::=--force-unsafe-io install --no-install-recommends \
     ca-certificates \
-  && rm -rf /var/lib/apt/lists/*
+  && apt-get clean \
+  && find /var/lib/apt/lists -type f -delete
 
 # Copy in the 'package' directory from the build image.
-COPY --from=build /src/package /opt/portier-broker
+RUN mkdir -p /opt/portier-broker && tar -C /opt/portier-broker -xf /data.tar.gz && rm /data.tar.gz
```
stephank, line 59: I like this functionality, but wonder if there's not a simpler way to accomplish this. Do you think this could work:

1. Build stage
   - FROM debian
   - Build the broker
2. Data stage
   - FROM debian
   - Copy in lang, res, tmpl
   - If $data_url is set, rm the above, use curl + tar
   - rm *.po
3. Final stage
   - FROM debian
   - Setup users, install deps
   - Copy in broker from build stage
   - Copy in data from data stage
   - Image settings stuff

What I'm hoping to accomplish here:
 - In the final stage, I originally tried to keep deletions to a minimum. The final stage is what people download, and with your changes, there's a hidden data.tar.gz download in an upper layer, which a lower layer deletes. (This could be a petty amount of bytes, then again, someone could slam their data file full with branding. 🙃)
 - I believe changing data also reruns apt? Because the order makes Docker think apt commands have a dependance on data.
 - I think we can just be crude in stage 2, and don't need a switch. (Though the switch IS neat. 😄) Even if we copy in local data, then delete it again, to then rebuild it from $data_url, it doesn't matter because only the final stage dictates what layers are in the output. Only downside is we can't do a conditional `ADD`, so that's why curl + tar.

jimdigriz: Sure this works for me, not really bothered about how it gets done, just that the result is I do not need to fork the project to just add my own branding.

You want me to respin the Dockerfile along these lines or did you want to? I am not opinioned on the implementation side so if you are more passionate about it, you will not get any complaints from me!

stephank:
@jimdigriz I'd like to merge this. Is it still WIP in your opinion?

jimdigriz:
I've had this in production for a client for a few months now and it has not needed to make any major changes since the end of October; had to move to a single region (instead of multi-region) as in production I found the C2S geo-targetting hit a different region to the S2S communication and so the Redis server in that region had no idea what was going on.

Everything has been working fine though since. I think it is ready to merge.

PR #253, opened 2021-06-12
Build releases for more platforms, multiarch Docker images

stephank:
I got a mail saying Docker autobuilds are going away, so we need to build and push from somewhere else, naturally GitHub Actions. I took the opportunity to work on improving the release scripts. This now builds for a larger set of platforms. The full set is now:

- Linux: ARMv6, ARMv7, ARM64 (v8), Intel 32-bit & 64-bit
- Windows: Intel 32-bit & 64-bit
- macOS: Apple Silicon, Intel 64-bit

See a test release here: https://github.com/portier/portier-broker/releases/tag/v0.4.0-test

This also pushes the Linux builds as a multiarch Docker image to Docker Hub. See the test release: https://hub.docker.com/repository/registry-1.docker.io/portier/broker/tags?name=v0.4.0-test

Here's the CI run of the test release: https://github.com/portier/portier-broker/actions/runs/931453513

### DIFF
--- a/.github/actions/prep-linux-cross/action.yml
+++ b/.github/actions/prep-linux-cross/action.yml
@@ -1,0 +1,12 @@
+name: prep-linux-cross
+description: Preparation for Linux cross compiling
+inputs:
+  targets:
+    description: Target triples to prepare for.
+    required: true
+  packages:
+    description: Debian packages to install for all targets.
+    required: false
+runs:
+  using: node12
+  main: main.js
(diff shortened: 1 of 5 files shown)


stephank:
I modified the macOS build to create a universal binary, but didn't really test it, because testing this is quite tedious. I did test the `lipo` command locally, so expect it to just work. 🤷‍♂️ If it really breaks, I'll fix it on the next actual release.

stephank:
Looks like the ARMv6 and ARMv7 builds don't actually work.

```
/lib/ld-linux-armhf.so.3: No such file or directory
```

Looking at the Debian base images, I think they're built for software floating point? And I was building for hardfloat.

I don't really know what's expected on these platforms, so I'm tempted to just remove them until someone comes along that actually needs them. :)

stephank:
Going to move this ahead without ARM 32-bit.

PR #270, opened 2021-07-16
Implement authorization code flow

stephank:
This implements the OAuth2 authorization code flow (`response_type=code`), which in practice is just an extra indirection for us. Instead of `id_token`, we return `code`, and a server-to-server request to our `/token` endpoint is used to redeem `code` for `id_token`.

Normally, `/token` is supposed to return an access token, but we simply set that to `"UNUSED"`. Given that this is for integration with generic OIDC clients, there's nothing an access token can be used for any way. We don't provide a `userinfo_endpoint` endpoint, for example.

In addition, this implements `response_mode=query`. We recommend against this, because in the implicit flow, a token may leak in ~~browser history~~ or server logs, but it is the default mode for `response_type=code` and required by many generic clients. In practice, this is a less of an issue for the authorization code flow, because the code is single-use. (**EDIT**: Thinking about this, `response_mode=fragment` could end up in browser history as well. But I think it's a small issue, because `nonce` also makes the token single-use.)

I also added a test to our E2E test suite, documented two integrations (OAuth2 Proxy and KeyCloak), and made various small fixes along the way.

One notable fix is how we handle the `scope` parameter. Per OIDC Core, we're supposed to require `openid` only, and ignore extra unknown values.


onli:
Would you recommend changing portier clients to use that new flow?

stephank:
> Would you recommend changing portier clients to use that new flow?

I don't see any benefit. It'd 'hide' the token from the user agent, but there's nothing confidential in there. It's just the email address with a signature.

PR #608, opened 2022-10-20
Improve email parsing for Lettre compatibility

stephank:
This adds validation using the `email_address` crate, which is what Lettre uses. This is on top of our existing validation. The crate has some additional checks that we don't do, like length checks on the local part and the domain part. Some of the simpler checks do overlap, so we delegate those now.

I suspect we were running into these missing checks later on when sending mail, based on reports from @tschaume. But at that point, we `.expect()` and crash.

In addition, it looks like we were converting our `EmailAddress` to Lettres`Mailbox` when sending, which also allows full name in the input string. We now instead convert to `Address`, hopefully eliminating any potential issues there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -970,6 +970,7 @@ dependencies = [
  "combine",
  "ctrlc",
  "docopt",
+ "email_address",
  "envy",
  "futures-util",
  "gettext",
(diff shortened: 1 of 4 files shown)


tschaume:
Wonderful @stephank! Thanks for the lightening quick turnaround including release and docker image. You're helping [Materials Project](https://materialsproject.org/) make materials science data accessible to people in the world who don't have accounts with the standard OAuth providers 👍

stephank:
Very cool! Thanks for reporting the issue, and hope this fixes it. 🙂

Issue #758, opened 2023-05-22
trim whitespace on paste of code


jimdigriz:
Could you add a [`onpaste`](https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event) to [`tmpl/confirm_email.mustache`](https://github.com/portier/portier-broker/blob/main/tmpl/confirm_email.mustache) to handle lesser {web,e}mail clients that seem to place sometimes considerable whitespace before the confirmation code?

Though you do remove all whitespace in [`src/bridges/email.rs:confirmation`](https://github.com/portier/portier-broker/blob/main/src/bridges/email.rs) I am seeing the problem before we get to the endpoint due to the text input field having a limit of 20 and it submits a truncated confirmation instead.

"Why does this happen, surely 20 is enough?", you may ask, and it is true and it is a good question...but it turns out software hates all people...

What I have observed:

1. the code emailed to me is `123456 abcdef` (12 characters + 1 space)
2. I double click on the code to paste it in to the confirmation box
3. what is actually pasted is N leading spaces, I see mostly 1 which works fine in the 20 character limit but I am able to regularly have it paste in 8 leading whitespace characters which results in the trailing `f` being truncated/ignored
     * you can see this yourself if you press the space bar eight times before pasting in the code
     * the HTML in the email template is just the paragraph element wrapping the code so it seemingly is not something I can fix there: ![image](https://github.com/portier/portier-broker/assets/2567406/f2f8ea4f-f1fc-4362-9ef6-33ac62aa0192)

Maybe we could include something like:
```
<script>
 document.querySelector("input[name='code']").addEventListener("paste", function(e) {
  e.preventDefault();
  this.value = e.clipboardData.getData("text/plain").trim()
 });
</script>
```

The reason I am proposing the `onpaste` hook is:
1. users may post trim the whitespace manually after seeing it pasted incorrectly but they are unlikely to notice a single character being trimmed off
2. the user can still pound the enter key blindly :)
3. I suspect we do not want to avoid the problem by just bumping the limit to something higher than 20, though this could work too? Maybe do both?

stephank:
Fixed in #781

PR #781, opened 2023-08-16
Support `_login_hint` to actually provide a hint

jimdigriz:
There are occasions where you want to provide `login_hint` but not have Portier auto submit the request and start the authentication for you.

My scenario is:
 * use Portier as the login landing page
 * application wants to auto-fill in the email address field with the last used email address
 * user has multiple hats (email addresses and thus authorised for different things) at an organisation so may wish to use a different one than they did last time

This PR does the following:
 1. make all parameters passed prepended with `_` (underscore) for private use by Portier
 2. if `_login_hint` is set, this is displayed in the `<input name="login_hint" ...>` form field
      * ~~annoyingly we cannot automatically move the cursor to the end without JS (and amending the HTML documents CSP) but I am okay with this~~ just added some JS to do this
 3. whilst in here, opportunistically update the input field from type `text` to [`email` to provide client side validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email)

This is the first time I have ever touched Rust, so forgive me for the ugly conditional or two in here :)

Not a problem if you wish to close the PR as this is something you do not want to accept, or for you to cherry-pick or take inspiration from to roll your own better version.

### DIFF
--- a/res/static/confirm_email.js
+++ b/res/static/confirm_email.js
@@ -1,0 +1,4 @@
+document.querySelector('input[name="code"]').addEventListener('paste', function(ev) {
+  ev.preventDefault();
+  this.value = ev.clipboardData.getData('text/plain').trim()
+});
--- a/res/static/login_hint.js
+++ b/res/static/login_hint.js
@@ -1,0 +1,7 @@
+document.addEventListener('DOMContentLoaded', function() {
+  // move the cursor to the end of the value
+  const input = document.querySelector('input[name="login_hint"]');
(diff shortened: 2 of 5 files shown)


jimdigriz:
...and cheekily slip in a fix for #758 but happy to break this out into another PR if needed.

stephank (approved):
This looks great! No objections. 👍

stephank:
Released in v0.8.4.

PR #906, opened 2024-07-22
French translation

daniel-g-gagnon:
I added a French translation and two locales, fr and fr_CA.

I could add just one, but both disagree heavily on one word and that word is email, so I can’t have a neutral translation that works for both.

The other difference is that fr has a narrow non-breaking space before a colon as favored by the Académie de la langue française, and fr_CA has a regular non-breaking space as favored by the Office québécois de la langue française. The two translations are otherwise identical.


stephank:
Thanks, also for the attention to detail! I release [v0.11.0](https://github.com/portier/portier-broker/releases/tag/v0.11.0) with this. 🙂

PR #968, opened 2025-02-06
Fixes to the contributed GCP deployment

jimdigriz:
Another week, another product in the Google Graveyard, this time it is Container Registry. One of the commits in this PR migrates to Artifact Registry...probably will be here again next week as that is set to be phased out too...</rant> :-/

The other commit is in response to O365 sucking in other ways too, turns out their name servers are unreliable and *often* fail to return A/AAAA records so we make `public_ip` configurable in the deployment template; correcting what it always should have been.


stephank:
> Another week, another product in the Google Graveyard

Ouch. Thanks for maintaining this. I can at least relate and sympathize a little with the YAML-engineering going on here. 🥲

PR #970, opened 2025-02-08
Upgrade Redis, use RESP3 for pubsub

stephank:
This finally upgrades the `redis` crate and relies on its support for pubsub, instead of our own solution. We do now require RESP3, which means Redis >= 6.0 (or valkey), which is more than 4 years old by now.


stephank:
Tried using ConnectionManager instead of MultiplexedConnection, but the way it does reconnects is a little opaque. There appears to be no logging or events to catch at all. Only `PushInfo::Disconnection`, but that has no additional information about next retry, for example.

Sticking with MultiplexedConnection keeps behavior as it previously was: we crash if redis disappears.

PR #979, opened 2025-03-04
Update dependencies

stephank:
Rust 1.81 was 6 months ago, so we can now use it as MSRV. We have rusqlite (#962) and some indirect dependencies that require it.


stephank:
Looks like rusqlite 0.33 requires a super recent rust, so reverting that upgrade.

PR #996, opened 2025-05-08
CORS support

### LINKED ISSUE #198: CORS Support
I'm hosting my own broken and in order to streamline the experience I'd like the make the initial authentication request as an async request in the background, rather than navigating the browser to the auth endpoint.

Since there is already a list of allowed origins, would it make sense to add an option to enable CORS for those origins? If so, which endpoints do we want to enable CORS for, all endpoints or only some?

I am happy to look into this further and see how a PR goes

jimdigriz:
This PR adds CORS headers depending on the value of `allowed_origins`.

Excuse the awful Rust code, I am not a Rust programmer, but this seems to work for me.

Can this be reviewed and considered by someone who knows what they are doing? :)

The 'development improvement' commit might be interesting to you, spares me having to install Rust directly on my workstation and allows you to skip the download and building of the dependencies on re-builds.

Resolves #198

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,11 @@
 FROM rust:1-alpine AS build
 RUN apk add --no-cache build-base
 WORKDIR /build
+#### BEGIN faster development https://stackoverflow.com/a/58474618 ####
+COPY Cargo.toml Cargo.lock ./
+RUN echo 'fn main() {}' > dummy.rs && sed -i -e 's#src/main.rs#dummy.rs#' Cargo.toml
+#### END faster development ####
+RUN cargo build --release --locked
 COPY . .
 RUN cargo build --release --locked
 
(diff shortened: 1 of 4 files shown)


stephank:
Thanks for this! I made some small changes and published [v0.13.0](https://github.com/portier/portier-broker/releases/tag/v0.13.0).

Notably, I put this behind an option `cors_ttl`, which controls max-age and defaults to zero. CORS headers won't be set unless this is set to something other than zero.

> The 'development improvement' commit might be interesting to you, spares me having to install Rust directly on my workstation and allows you to skip the download and building of the dependencies on re-builds.

This is great! I did swap it for the second solution from StackOverflow, because I believe that should work all-around better. (Doesn't create a ~800MB layer containing build artifacts, and Cargo should be able to do partial rebuilds even on `Cargo.toml` / `Cargo.lock` changes.) Let me know if it's no good for your case. 🙂

You can also add `--build-arg cargo_flags=""` to docker build now to get debug builds, which should also be much faster.

PR #1007, opened 2025-05-30
Add device confirmation to thwart email scanners

### LINKED ISSUE #1005: Email scanners following the confirmation link
We're receiving reports of Microsoft scanners following confirmation email links, executing JS, and making the POST request. This renders the confirmation link useless to users, and leaks RP credentials to whatever sandbox Microsoft scanners are using, even if those are probably wiped afterwards.

Looking around, it seems we're not alone: https://berthub.eu/articles/posts/shifting-cyber-norms-microsoft-post/

I can think of 2 solutions:

1. Eliminate the confirmation link in emails, and require the user type out the code.
2. Add a manual step at the confirmation link, where it shows a page with a big 'continue' button.

I prefer 1, because I feel like 2 is surprising/confusing and maybe just as temporary if scanners become even more annoying. But maybe there are other ideas?

Pinging some active users: @onli @jimdigriz

stephank:
Fixes #1005
Replaces #1006

Looks like this in dutch:

Screenshot 2025-05-30 at 16 38 36


stephank:
Hoping for a quick check of the translations: 🙂
- @onli, German
- @daniel-g-gagnon, French

jimdigriz:
This is opens a legal element. To do this you will need potentially a cookie consent modal and definately a link to a privacy policy.

I would recommend just adding instead a simple '1+1' captcha box or ask the user to re-enter their email address.

jimdigriz:
The problem is that where one needs to exist you must run this through your own legal council to figure out if you can use this software.

If we store nothing but a session cookie; or something that lives no longer than the date of the expiry of when the session has to be complete by, there would be no need for this.

My clients have no problem with this, as they already have the 'evil' cookies all over their websites, but for my own stuff, I go out my way to use cookie-less free analytics so I can actually avoid a cookie consent banner as the web is already a terrible place...I needn't make it worse :)

